PR #57, opened 2018-12-01
Fixed indentation issue

skinnylatte:
When there is a code block nestled within a numbered list

And / or

A second / third paragraph

Indentation breaks.

This aims to fix that issue with proper GitHub-flavored Markdown indentation.


skinnylatte:
mostly works but there was an issue with no 6. fixing that to resubmit.

PR #113, opened 2018-12-10
Inaccuracies fix - 2

skinnylatte:
* Front page text on connection pooling
* Vitess vs. Vanilla MySQL Table
* Server config page edits


skinnylatte:
second batch of inaccuracy fixes.

PR #122, opened 2018-12-15
Update local.md

deepthi:
Direct people to slack rather than to google groups.


deepthi:
I'm going to do this from the CLI so that I can signoff the commit properly.

PR #144, opened 2019-01-28
Fix escaping of instructions

dasl-:
This escaping issue causes the indentation of the docs to be funky: https://vitess.io/docs/launching/server-configuration/#initialization


deepthi:
@dasl- the change looks good. Can you fix the DCO?

dasl-:
@deepthi fixed!

PR #175, opened 2019-03-29
fixed getting-started links

toravir:
i don't have hugo installed on my dev machine - but did my best judgement in coming up with the changes..


toravir:
Signed-off-by: Ravi R <ravir@employees.org>

PR #187, opened 2019-05-06
Remove symlink reference

gedgar:
Signed-off-by: Gary Edgar <gary@planetscale.com>

### DIFF
--- a/content/docs/tutorials/local.md
+++ b/content/docs/tutorials/local.md
@@ -135,7 +135,7 @@ In addition, Vitess requires the software and libraries listed below.
 
 ### Build Vitess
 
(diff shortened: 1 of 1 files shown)


deepthi:
> @deepthi I saw that you added this symlink previously, which I just removed. AFAIK, go doesn't like to build on files that are sym-linked. Was there a specific use case where you had to do this, and that it worked?

where are you seeing that? I don't see it in the file history. I recall @dkhenry telling me to create a vitess.io/vitess directory under go/src instead of symlinks, and that is how I have my local env set up.

sougou:
This was in the old history in vitess: https://github.com/vitessio/vitess/commit/f9020f0e49708553ebe0c088665bdf5a6ba694f9

deepthi:
> This was in the old history in vitess: [vitessio/vitess@f9020f0](https://github.com/vitessio/vitess/commit/f9020f0e49708553ebe0c088665bdf5a6ba694f9)

We should be good to remove those now.

Issue #208, opened 2019-07-02
Documentation correction


PrismaPhonic:
I noticed what seems to be a typo in the documentation for the "Run Vitess Locally" tutorial.  It states the following:

> if you ran make build while in $HOME/vt/src/vitess.io/vitess, then you should set: export VTROOT=$HOME/vt

Then later the same tutorial has us run a `101_initial_cluster.sh` script which in turn eventually runs a `vttablet_up.sh` script.  In the `vttablet_up.sh` script we see references to VTROOT such as:

```
init_db_sql_file="$VTROOT/config/init_db.sql"
```

It seems based on this that VTROOT should be pointed directly at the vitess repo root folder, not to a folder that would precede src/vitess.io/vitess.  

Please correct me if I'm wrong, but it seems like the documentation should instead say:

```
if you ran make build while in $HOME/vt/src/vitess.io/vitess, then you should set: export VTROOT=$HOME/vt/src/vitess.io/vitess
```

If this seems right I would be happy to issue a PR with the change to correct this.

@avaidyanatha ?

PrismaPhonic:
I'm going to close this issue since the documentation is correct.  What was inconsistent with the bootup script is that config as a folder should get symlinked to VTROOT which resolves what looks like inconsistent uses of VTROOT.

PR #209, opened 2019-07-02
Fixed documentation so VTROOT example is correct

PrismaPhonic:
VTROOT example was not pointing at root repo as discussed in #208


PrismaPhonic:
Please don't merge this yet.  It seems as thought he documentation was correct but parts of the vttablet-up.sh script was wrong.  will fix

PR #216, opened 2019-07-19
Promote concepts to top level menu item

morgo:
This PR moves overview/concepts to a top-level menu item (partially addressing #212). It also splits the page into smaller pages, since I intend to link to the specific terminology from other parts of the manual.

I will handle some of the other changes in separate PRs. I am trying to break it up to make translation and review easier.

Edit: an alias has also been added so there should not be any broken links.


morgo:
@Johnny-Three PTAL

Let me know if anything doesn't make sense, or won't work when it is time for translation.

Johnny-Three:
LGTM

PR #236, opened 2019-07-30
Update Overview

morgo:
This moves architecture, history and cloud-native to top level sub-documents. The rationale being that the original document is too long, and splitting these out helps draw attention.

The architecture page content still duplicates "Concepts" -- an issue I will still have to resolve.

The page "scaling with Vitess" has been removed from Overview. This is perhaps a controversial move - I'm looking for feedback. I felt it was too detailed for Overview - it was more like a task-based guide, but covered a lot of topics. It also duplicated content; for example the architecture diagram also shows up on the architecture page.


morgo:
Do not merge yet. I will ask a wider audience to review since it is dropping content.

lkburns:
This works for me. I found the section Scaling with Vitess useful b/c it gave a big picture view of what using Vitess might look like, but the content outdated and somewhat duplicative of the tutorials. At some point it would be nice to add in something give generalist the big picture view of Vitess that is missing in the tutorials and in the What is Vitess section.

morgo:
I checked with Sugu, and he's onboard with the changes. It is more in-line with how we present Vitess at conferences too.

PR #237, opened 2019-07-30
Cleanup Replication Documentation, rm schema-swap, update-stream, filtered replication

morgo:
This is a partial followup to https://github.com/vitessio/website/pull/220

Fixes https://github.com/vitessio/website/issues/241
Fixes https://github.com/vitessio/website/issues/238

The schema-swap should have been removed from more locations. There is also a lot of content around SBR, which is no longer recommended.

Update stream has also been removed, since it is replaced by vstream (still needs to be documented, but update stream should not be used).

<strike>I am sure there will be future edits (these 2 pages should merge), but the goal of this PR is just to remove what is out of date.</strike>

I've also merged the vitess replication and mysql row-based replication pages and called it "mysql replication" (to disambiguate from Vreplication). Inner workings of how RBR works are removed.


morgo:
> 1. When we recommend RBR we might want to strongly discourage SBR or enumerate what features won't work with SBR.

I am not sure we should enumerate, because this might still encourage usage.. which will put us in a bad position in the future as we have other plans :-) But I have toned up the language from "recommends" to only RBR is supported.

> 2. enable_semi_sync is not set to true by default unfortunately. we need to rewrite that section around this fact.

Fixed, thanks.

I changed the scope of this PR to merge the two replication documents. I removed some content that described how RBR works (it's not important) and the compare and contrast from SBR (also not important now that only one is supported).

morgo:
I've also renamed the page from "Vitess Replication" to "MySQL Replication" because otherwise the relationship to "VReplication" is a bit weird.

morgo:
I've also removed filtered replication, since vreplication should be used instead.

Issue #210, opened 2019-07-02
Rename default branch to 'prod'


morgo:
GitHub's [robots.txt](https://github.com/robots.txt) explicitly permits indexing the branch called 'master'. This is fine for source code, but for docs it creates a duplicate content problem in search engines.

It is also better to have Search Engines index https://vitess.io/docs/ instead of GitHub because we can add redirection if content needs to be moved around.

I suggest we rename the default branch to 'prod' to fix this.

@lucperkins is this something you mind helping with?

See also:
https://stackoverflow.com/questions/15844905/how-to-stop-google-indexing-my-github-repository

morgo:
This has been completed.

PR #248, opened 2019-08-02
translate overview part into chinese

Johnny-Three:
@morgo PTAL


morgo:
Just a quick note: the PR was against `master`. The branch changed to `prod` in https://github.com/vitessio/website/issues/210 -- I will push `master` into `prod` so this is in the correct branch.

PR #255, opened 2019-08-08
Remove MYSQL_FLAVOR requirement

morgo:
This is no longer required since
https://github.com/vitessio/vitess/pull/4987 merged

This is currently accurate for the build from source and vagrant tutorials, but the local tutorial needs to wait until a new [vitess-release](https://github.com/planetscale/vitess-releases/releases) is built. 

cc @gedgar 


gedgar:
@morgo, Running `yum install mysql-server` yields `No package mysql-server available.` on a fresh CentOS 7 box. I'm assuming this is because MariaDB 5.5 replaced MySQL in CentOS 7, and MariaDB 10.3 replaces MySQL in RHEL 8. Do we want to update the MySQL installation instructions for RHEL/CentOS?

Maybe something like this? 
```
sudo yum localinstall https://dev.mysql.com/get/mysql57-community-release-el7-9.noarch.rpm
sudo yum install mysql-community-server
```

morgo:
Thanks @gedgar - I've changed the yum steps.

The apt steps I've left the same, since they are correct for Ubuntu. But for Debian there is a similar deal, where it is now MariaDB. We could do something like this, but I think I'm going to kick that can down the road, since it is still correct for a lot of people. What do you think?

```
wget https://dev.mysql.com/get/mysql-apt-config_0.8.13-1_all.deb
sudo apt install -y ./mysql-apt-config_0.8*
sudo sed -i 's/mysql-8.0/mysql-5.7/' /etc/apt/sources.list.d/mysql.list
sudo apt-get update
sudo apt-get install mysql-community-server mysql-community-client
```

morgo:
I will address this in https://github.com/vitessio/website/pull/323 and https://github.com/vitessio/website/pull/309 instead. Closing.

PR #262, opened 2019-08-10
Move launching and schema management content

morgo:
Fixes https://github.com/vitessio/website/issues/212 !

This moves the content from launching and schema management to other sections of the docs. Most of the content stays the same, but I have edited the scalability philosophy for brevity since it is now an "overview". I think the content works really well in explaining why Vitess does things a certain way (usually how we build or why we built is a blog post, but I think it works well here too.)

It is the last of TOC refactoring PRs. From here the focus will be on improving the user guide and reference sections.


content/en/docs/overview/scalability-philosophy.md
```diff
@@ -0,0 +1,75 @@
...
+
+* Scalability: There are situations where multi-master gives you a little bit of additional runway. However, since the statements have to eventually be applied to all masters, it’s not a sustainable strategy. Vitess addresses this problem through sharding, which can scale indefinitely.
+* High availability: Vitess integrates with Orchestrator, which is capable of performing a failover to a new master within seconds of failure detection. This is usually sufficient for most applications.
+* Low-latency geographically distributed writes: This is one case that is not addressed by Vitess. The current recommendation is to absorb the latency cost of long-distance round-trips for writes. If the data distribution allows, you still have the option of sharding based on geographic affinity. You can then setup masters for different shards to be in different geographic location. This way, most of the master writes can still be local.
+
+## Multi-cell
+
+Vitess is meant to run in multiple data centers / regions / cells. In this part, we'll use cell as a set of servers that are very close together, and share the same regional availability.
```
deepthi:
```suggestion
Vitess is meant to run in multiple data centers / regions / cells. In this part, we'll use "cell" to mean a set of servers that are very close together, and share the same regional availability.
```

content/en/docs/overview/scalability-philosophy.md
```diff
@@ -0,0 +1,75 @@
...
+
+A cell typically contains a set of tablets, a vtgate pool, and app servers that use the Vitess cluster. With Vitess, all components can be configured and brought up as needed:
+
+* The master for a shard can be in any cell. If cross-cell master access is required, vtgate can be configured to do so easily (by passing the cell that contains the master as a cell to watch).
+* It is not uncommon to have the cells that can contain the master be more provisioned than read-only serving cells. These *master-capable* cells may need one more replica to handle a possible failover, while still maintaining the same replica serving capacity.
```
deepthi:
```suggestion
* It is not uncommon to have cells that can contain the master be more provisioned than read-only serving cells. These *master-capable* cells may need one more replica to handle a possible failover, while still maintaining the same replica serving capacity.
```

content/en/docs/overview/scalability-philosophy.md
```diff
@@ -0,0 +1,75 @@
...
+* Failing over from one master in one cell to a master in a different cell is no different than a local failover. It has an implication on traffic and latency, but if the application traffic also gets re-directed to the new cell, the end result is stable.
```
deepthi:
```suggestion
* Failing over from a master in one cell to a master in a different cell is no different than a local failover. It has an implication on traffic and latency, but if the application traffic also gets re-directed to the new cell, the end result is stable.
```

content/en/docs/overview/scalability-philosophy.md
```diff
@@ -0,0 +1,75 @@
...
+* It is also possible to have some shards with a master in one cell, and some other shards with their master in another cell. vtgate will just route the traffic to the right place, incurring extra latency cost only on the remote access. For instance, creating U.S. user records in a database with masters in the U.S. and European user records in a database with masters in Europe is easy to do. Replicas can exist in every cell anyway, and serve the replica traffic quickly.
+* Replica serving cells are a good compromise to reduce user-visible latency: they only contain replica servers, and master access is always done remotely. If the application profile is mostly reads, this works really well.
+* Not all cells need `rdonly` (or batch) instances. Only the cells that run batch jobs, or MapReduce jobs, really need them.
```
deepthi: remove reference to MapReduce? and add reference to OLAP?

PR #278, opened 2019-08-23
Add blog post draft on 250g shards

morgo:
The images are a bit large, I think I need to edit the css theme to fix it. I gave @raunaqgupta a ping to ask for help :-)

Content is ready for review though. Looking forward to your feedback.


content/en/blog/2019-08-23-why-250gb-shards.md
```diff
@@ -0,0 +1,145 @@
...
+
+### How far can an efficient inserts scale?
+
+Table A only lost 13% of its insert performance when the buffer pool was lowered from 16GB to 128MB. To prove that there is not a clear “maximum number of rows” limit, lets now extend the test run time to 5 hours. After inserting almost 463 million rows, we can see our 376GB table still retains most of its insert performance:
+
+![Table A insert over 5 hours](/files/250g-bench/image5.png)
+
+Performance is sustained while inserting for 5 hours. At 463 million rows, there is only an 18% drop from the 113 million rows inserted in 1 hour. Internally InnoDB uses pages to store, and the buffer pool cache is page oriented. There is no immediate supporting evidence that there is a row-count limit on table size.
+
+**TL;DR: Insert performance is not limited by data size, or number of rows. It is dependent on the table+index structure, and how rows are inserted. It is really difficult to provide a generalized answer here. You could have a 256GB database that works well with 1GB RAM, and another 256GB database that requires 128GB RAM.**
```
sougou: Will it be possible to simplify this by counting size of all data, including indexes? Essentially, it would be equivalent to inserting to two tables. Table B has higher cost because the 'second' table is bigger. If results are consistent, then the only differentiator would be  whether the values are monotonically increasing or not.

Let's say:
R: row size
B: bytes per second
E: insert efficiency

For table A, the insert rate would be: (B/R(a))*E+ (B/R(b_idx))*E
For B: (B/R(a))*E + (B/R(c_idx))

This is just a comment, and will probably need a separate blog post.

sougou (approved):

content/en/blog/2019-08-23-why-250gb-shards.md
morgo: You are right, I never put Table A and Table B on the same chart intentionally because of the secondary indexes being different. But I could normalize it and compare if I evaluated the size of the table.

It is not completely reliable because there is some pre-allocation done, and I think the non-insert efficient probably has a lower page-fill factor (don't want to let users believe larger == better), but a line graph over time may be interesting.

I will have to re-run to capture the data. I'll see if it turns out useful.

sougou: The main idea is about being able to get a ballpark estimate of the cost and throughput based on the table structure. So, if someone says they're getting X throughput, we can tell if they're in the ballpark or way off.

This diverges from the blog topic, which is why I thought it would be a separate blog post.

content/en/blog/2019-08-23-why-250gb-shards.md
```diff
@@ -0,0 +1,156 @@
+---
+author: 'Morgan Tocker'
+date: 2019-08-23T09:07:21-08:00
+slug: '2019-08-23-why-250gb-shards'
+tags: ['Guides']
+title: 'Why does Vitess recommend 250GB per MySQL server?'
+---
+
+Vitess has an [opinionated approach](https://vitess.io/docs/overview/scalability-philosophy/) to database scalability. Some of those opinions have minimal controversy such as how durability should be provided via replication. But the one I find interesting is the 250GB per MySQL server recommendation.
```
avaidyanatha:
```suggestion
Vitess has an [opinionated approach](https://vitess.io/docs/overview/scalability-philosophy/) to database scalability. Some of those opinions have minimal controversy such as how durability should be provided via replication, but the one I find interesting is the 250GB per MySQL server recommendation.
```

content/en/blog/2019-08-23-why-250gb-shards.md
```diff
@@ -0,0 +1,156 @@
...
+- When the master fails; the blast radius will be much larger, potentially taking more of the service offline.
+- When a replica fails; it will likely take 10-12 hours to restore a new replica at gigabit network speed. While this time can be reduced with a higher speed network (recommended); there is a fairly large window for a secondary failure could occur. Many Vitess users aim for 15 minute recoveries; which is possible with a 250G shard on a 2.5Gbps network.
+
+Vitess **does not** impose 250GB as a hard limit. It even encourages you to run multiple instances of MySQL (multiple tablets) on the same host.
+
+**TL;DR: Vitess suggests 250G as a shard size for management and predictable recovery times. It may be true for performance reasons too, but that depends a lot on your usage.**
+
+## Conclusion
+
+By specifying a recommended size, the authors of Vitess can also make assumptions about how long certain operations should take and simplify the design of the system. This loops back to my opening sentence: Vitess has an opinionated approach to scalability. Not attempting to be all things to all people is a good thing :-)
```
avaidyanatha: The conclusion can be a little more beefy and should reiterate the statements about how there is no strict physical or "one-size-fits-all" limit for MySQL.

content/en/blog/2019-08-23-why-250gb-shards.md
```diff
@@ -0,0 +1,156 @@
...
+- Reducing replica delay
+
+**Let's use a 4TB shard failure as an example:**
+
+- When the master fails; the blast radius will be much larger, potentially taking more of the service offline.
+- When a replica fails; it will likely take 10-12 hours to restore a new replica at gigabit network speed. While this time can be reduced with a higher speed network (recommended); there is a fairly large window for a secondary failure could occur. Many Vitess users aim for 15 minute recoveries; which is possible with a 250G shard on a 2.5Gbps network.
+
+Vitess **does not** impose 250GB as a hard limit. It even encourages you to run multiple instances of MySQL (multiple tablets) on the same host.
+
+**TL;DR: Vitess suggests 250G as a shard size for management and predictable recovery times. It may be true for performance reasons too, but that depends a lot on your usage.**
```
avaidyanatha: TL;DR should go either at the beginning or after the conclusion :)

content/en/blog/2019-08-23-why-250gb-shards.md
```diff
@@ -0,0 +1,156 @@
...
+By prepending the random bytes with 26 bytes of a timestamp, the value becomes insert efficient. With our 16GB buffer pool we now have a very low rate of read-modify-write and can sustain performance as the table grows.
+
+<figure>
+    <img src="/files/250g-bench/image1.png"/>
+    <figcaption>Comparing inserts per second performance between “insert efficient” and random values for c. Higher is better. Notice how insert efficient sustains performance as the table becomes larger.</figcaption>
+</figure>
+
+### How far can an efficient inserts scale?
+
+Table A only lost 13% of its insert performance when the buffer pool was lowered from 16GB to 128MB. To prove that there is not a clear “maximum number of rows” limit, lets now extend the test run time to 5 hours. After inserting almost 463 million rows, we can see our 376GB table still retains most of its insert performance:
```
avaidyanatha:
```suggestion
Table A only lost 13% of its insert performance when the buffer pool was lowered from 16GB to 128MB. To prove that there is not a clear “maximum number of rows” limit, let's now extend the test run time to 5 hours. After inserting almost 463 million rows, we can see our 376GB table still retains most of its insert performance:
```

content/en/blog/2019-08-23-why-250gb-shards.md
```diff
@@ -0,0 +1,156 @@
...
+slug: '2019-08-23-why-250gb-shards'
+tags: ['Guides']
+title: 'Why does Vitess recommend 250GB per MySQL server?'
+---
+
+Vitess has an [opinionated approach](https://vitess.io/docs/overview/scalability-philosophy/) to database scalability. Some of those opinions have minimal controversy such as how durability should be provided via replication. But the one I find interesting is the 250GB per MySQL server recommendation.
+
+## Is this a physical MySQL Limit?
+
+**In short: no**. By “physical limit” I mean is there a file format restriction that says databases can not be greater than 250GB.
```
avaidyanatha:
```suggestion
**In short: no**. By “physical limit” I mean is there a file format restriction that says databases can not be greater than 250GB?
```

content/en/blog/2019-08-23-why-250gb-shards.md
```diff
@@ -0,0 +1,156 @@
...
+```
+CREATE TABLE tableb (
+ id INT NOT NULL PRIMARY KEY auto_increment,
+ b DATETIME NOT NULL,
+ c VARBINARY(512) NOT NULL,
+ UNIQUE (c)
+);
+```
+
+.. they look almost identical right? Lets try and insert rows into them in 32 parallel threads for one hour:
```
avaidyanatha:
```suggestion
... They look almost identical right? Lets try and insert rows into them in 32 parallel threads for one hour.
```

content/en/blog/2019-08-23-why-250gb-shards.md
```diff
@@ -0,0 +1,156 @@
...
+
+```
+INSERT INTO {tablename} (b, c) VALUES (NOW(), RANDOM_BYTES(512)), (NOW(), RANDOM_BYTES(512)), ..;
+```
+
+<figure>
+    <img src="/files/250g-bench/image4.png"/>
+    <figcaption>In one hour, I was able to insert 113 million rows. The insert performance was fairly consistent despite the final table size (93GB) being much larger than the InnoDB Buffer Pool (16GB). I suspect the initial sharp drop in performance is the SSDs write cache being exceeded.</figcaption>
```
avaidyanatha:
```suggestion
    <figcaption>In one hour, I was able to insert 113 million rows. The insert performance was fairly consistent despite the final table size (93GB) being much larger than the InnoDB Buffer Pool (16GB). I suspect the initial sharp drop in performance is the SSD's write cache being exceeded.</figcaption>
```

content/en/blog/2019-08-23-why-250gb-shards.md
```diff
@@ -0,0 +1,156 @@
...
+</figure>
+
+<figure>
+    <img src="/files/250g-bench/image6.png"/>
+    <figcaption>Insert performance dropped as the table got larger. The 16GB buffer pool was not enough to hold all the important pages, and iostat showed a high number of reads/s as pages needed to be read-modify-written. The final number of inserted rows was 29 million, with a 50GB table size.
```
avaidyanatha:
```suggestion
    <figcaption>Insert performance dropped as the table got larger. The 16GB buffer pool was not enough to hold all the important pages, and iostat showed a high number of reads per second as pages needed to be read-modify-written. The final number of inserted rows was 29 million, with a 50GB table size.
```

content/en/blog/2019-08-23-why-250gb-shards.md
```diff
@@ -0,0 +1,156 @@
...
+</figcaption>
+</figure>
+
+### Why do these two tables perform so wildly different?
+
+The index on c is wider (512 bytes vs. 6 bytes), but the random insertion combined with uniqueness is the real performance killer over time.
+
+With Table A, the performance is stable and consistent because the inserts are to the end of the table and the required pages are in memory. I repeated the benchmark with a 128MB buffer pool, and performance was only 13% different:
```
avaidyanatha:
```suggestion
With Table A, the performance is stable and consistent because the inserts are to the end of the table and the required pages are in memory. I repeated the benchmark with a 128MB buffer pool and observed only a 13% decrease in performance:
```

content/en/blog/2019-08-23-why-250gb-shards.md
```diff
@@ -0,0 +1,156 @@
...
+1 row in set (0.00 sec)
+```
+
+### Converting to Insert Efficient
+
+By prepending the random bytes with 26 bytes of a timestamp, the value becomes insert efficient. With our 16GB buffer pool we now have a very low rate of read-modify-write and can sustain performance as the table grows.
+
+<figure>
+    <img src="/files/250g-bench/image1.png"/>
+    <figcaption>Comparing inserts per second performance between “insert efficient” and random values for c. Higher is better. Notice how insert efficient sustains performance as the table becomes larger.</figcaption>
```
avaidyanatha: Style-wise, analysis or comments about a picture are generally left outside the caption, but I don't have a strong feeling about it one way or the other.

content/en/blog/2019-08-23-why-250gb-shards.md
```diff
@@ -0,0 +1,156 @@
...
+### How far can an efficient inserts scale?
+
+Table A only lost 13% of its insert performance when the buffer pool was lowered from 16GB to 128MB. To prove that there is not a clear “maximum number of rows” limit, lets now extend the test run time to 5 hours. After inserting almost 463 million rows, we can see our 376GB table still retains most of its insert performance:
+
+<figure>
+    <img src="/files/250g-bench/image5.png"/>
+    <figcaption>Performance is sustained while inserting for 5 hours. At 463 million rows, there is only an 18% drop from the 113 million rows inserted in 1 hour. Internally InnoDB uses pages to store, and the buffer pool cache is page oriented. There is no immediate supporting evidence that there is a row-count limit on table size.</figcaption>
+</figure>
+
+**TL;DR: Insert performance is not limited by data size, or number of rows. It is dependent on the table+index structure, and how rows are inserted. It is really difficult to provide a generalized answer here. You could have a 256GB database that works well with 1GB RAM, and another 256GB database that requires 128GB RAM.**
```
avaidyanatha: We probably shouldn't have two TL;DR sections.

content/en/blog/2019-08-23-why-250gb-shards.md
```diff
@@ -0,0 +1,156 @@
...
+
+- Taking a full backup
+- Provisioning a new read replica
+- Restoring a backup
+- Making a schema change
+- Reducing replica delay
+
+**Let's use a 4TB shard failure as an example:**
+
+- When the master fails; the blast radius will be much larger, potentially taking more of the service offline.
```
avaidyanatha:
```suggestion
- When the master fails, the blast radius will be much larger, potentially taking more of the service offline.
```

content/en/blog/2019-08-23-why-250gb-shards.md
```diff
@@ -0,0 +1,156 @@
...
+- When a replica fails; it will likely take 10-12 hours to restore a new replica at gigabit network speed. While this time can be reduced with a higher speed network (recommended); there is a fairly large window for a secondary failure could occur. Many Vitess users aim for 15 minute recoveries; which is possible with a 250G shard on a 2.5Gbps network.
```
avaidyanatha:
```suggestion
- When a replica fails, it will likely take 10-12 hours to restore a new replica at gigabit network speed. While this time can be reduced with a higher speed network (recommended); there is a fairly large window for a secondary failure could occur. Many Vitess users aim for 15 minute recoveries; which is possible with a 250G shard on a 2.5Gbps network.
```

content/en/blog/2019-08-23-why-250gb-shards.md
morgo: Watching size (data+index) grow is fascinating:
image1

Initially Table B is faster. I don't know if it's page splits growing space faster, or that the index values are large - and serializing large values into pages is more efficient than small values?

It's too complicated for now, I'll keep it out of scope. To answer your question though, I think the formula needs to account for change buffering too - which makes it difficult to project rate.

content/en/blog/2019-08-23-why-250gb-shards.md
morgo: My opinion is that the caption should explain what to look for in the graphic. It's probably a bit long though.

morgo:
@raunaqgupta @avaidyanatha Changes made. I've bumped the date to Monday. Assuming everything else is good, we can post then.

PR #281, opened 2019-09-03
Change video on homepage

morgo:
Fix bug with aliases
Rename "Manual Build" to "Local Install"

### DIFF
--- a/layouts/partials/home/video.html
+++ b/layouts/partials/home/video.html
@@ -1,14 +1,20 @@
 <section class="section is-light">
   <div class="container">
-    <div class="columns is-variable is-8 is-vcentered">
+    <div class="columns is-variable is-6 is-vcentered">
       <div class="column">
         <p class="title is-size-2 is-size-3-mobile has-text-weight-light">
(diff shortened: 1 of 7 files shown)


layouts/partials/home/video.html
```diff
@@ -8,7 +8,7 @@
       </div>
 
       <div class="column has-text-centered">
-        <iframe class="is-video" src="https://www.youtube.com/embed/q65TleTn2vg?rel=0" allowfullscreen="" width="480" height="270" frameborder="0"></iframe>
+        <iframe class="is-video" src="https://www.youtube.com/embed/E6H4bgJ3Z6c?rel=0" allowfullscreen="" width="480" height="270" frameborder="0"></iframe>
```
sougou: Would it be possible to fix the aspect ratio?

Screenshot from 2019-09-03 10-38-54

morgo: I think that is caused by CSS. I believe the height displayed is less than 270. I will ask @raunaqgupta to take a look in a followup PR.

sougou (approved):

PR #303, opened 2019-10-07
Update vreplication.md

deepthi:
`Stopping` is not a valid vreplication state, it should be documented correctly as `Stopped`.


morgo (approved):
LGTM

Issue #263, opened 2019-08-11
User guides are not all task oriented


morgo:
* VTTablet modes should be multiple Guides (Use Vitess with externally Managed MySQL etc)
* VTExplain should be something like "Check Queries for Compatibility"
* Backing Up and Restoring Data is fine (although "Backup and Restore" is shorter)
* Reparenting: Need to think about it, might be multiple guides.
* Topology Service: This is a reference, not a guide
* Transparent Securit Model: This is also a reference
* Upgrading: This is good how it is
* Schema management: might need to be split with some moving to reference. A task guide might be "making schema changes".
* Server configuration: might be something like "Configuring Vitess Components"

PR #309, opened 2019-10-18
Initial split of get started -> local into user guides

### LINKED ISSUE #298: Update local tutorial (Run Vitess locally)
Local example now uses etcd, but doc still has zookeeper.
Review and fix references to the example business. There is no introduction, and just before resharding, we start talking about yerba mate.

### LINKED ISSUE #315: Investigate Centos 7 get started tutorial
It looks like there might be some issues (Oracle Linux used here for context):
https://github.com/gaogaolulu/how_to_run_vitess_locally

morgo:
This is a documentation refactor. Let me explain the motivations:

- The current user-guides are not very task-oriented (https://github.com/vitessio/website/issues/263). User guide should mean "here is a set of steps to achieve a specific goal".
- Part of this is due to the use of shell scripts like `./301_customer_sharded.sh`. i.e. there is no expressive API that is truly universal (context is key, including working directory).  We should be making this script obsolete. i.e. instead of:
```
set -e

./lvtctl.sh ApplySchema -sql-file create_commerce_seq.sql commerce
./lvtctl.sh ApplyVSchema -vschema_file vschema_commerce_seq.json commerce
./lvtctl.sh ApplySchema -sql-file create_customer_sharded.sql customer
./lvtctl.sh ApplyVSchema -vschema_file vschema_customer_sharded.json customer

disown -a
```

.. we make sure that `vtctlclient` accepts a config file (issue https://github.com/vitessio/vitess/issues/4921) and replace the script with a manual page that shows a solution that works for all deployment methods:
```
vtctlclient ApplySchema -sql-file create_commerce_seq.sql commerce
vtctlclient ApplyVSchema -vschema_file vschema_commerce_seq.json commerce

vtctlclient ApplySchema -sql-file create_customer_sharded.sql customer
vtctlclient ApplyVSchema -vschema_file vschema_customer_sharded.json customer
```

If we do this, the examples also become more portable.. since I can figure out how to apply them to my use case without copying the whole examples directory.

----

So now to describe what is changed here:

- The local guide only covers deployment, and then suggests a next step.
- The next step of Vertical split uses MySQL from a global location (because the local guide showed how to setup mysql cli config).
- The user guides for vertical split and horizontal split are closer to deployment scenario agnostic (closer but not there yet. The goal should be that one could go from the Kubernetes or Vagrant deploy guide straight to here too, with one version to maintain.)

Also:
Fixes #298 
Fixes #315

Long term this will also lead to less duplicate content, since there is overlap between kubernetes and local. It will also make it easier to add deploy guides for every cloud deployment scenario separately (GKE, EKS, Azure, etc.) without the duplicate content.

### DIFF
--- a/content/en/docs/get-started/local.md
+++ b/content/en/docs/get-started/local.md
@@ -6,7 +6,7 @@ featured: true
 aliases: ['/docs/tutorials/local/']
 ---
 
-This guide covers installing Vitess locally for testing purposes, from pre-compiled binaries. We will launch 3 copies of `mysqld` initially, and then many more as you work through the tutorial, so it is recommended to have greater than 4GB RAM, as well as 20GB of available disk space.
+This guide covers installing Vitess locally for testing purposes, from pre-compiled binaries. We will launch multiple copies of `mysqld`, so it is recommended to have greater than 4GB RAM, as well as 20GB of available disk space.
 
 ## Install Packages
 
@@ -18,14 +18,15 @@ PlanetScale provides [weekly builds](https://github.com/planetscale/vitess-relea
 # Apt based
 sudo apt-get install mysql-server
 # Yum based
-sudo yum install mysql-server
+sudo yum localinstall https://dev.mysql.com/get/mysql57-community-release-el7-9.noarch.rpm
+sudo yum install mysql-community-server
 ```
 
 _Vitess supports MySQL 5.6+ and MariaDB 10.0+. We recommend MySQL 5.7 if your installation method provides a choice._
 
 ## Disable AppArmor
 
-We recommend that you uninstall or disable AppArmor. Some versions of MySQL come with default AppArmor configurations that the Vitess tools don't yet recognize. This causes various permission failures when Vitess initializes MySQL instances through the mysqlctl tool. This is an issue only in test environments. If AppArmor is necessary in production, you can configure the MySQL instances appropriately without using `mysqlctl`:
+We recommend that you uninstall or disable AppArmor since it may causes permission failures when Vitess initializes MySQL instances through the `mysqlctl` tool. This is an issue only in test environments. If AppArmor is necessary in production, you can configure the MySQL instances appropriately without using `mysqlctl`:
 
 ```bash
 sudo service apparmor stop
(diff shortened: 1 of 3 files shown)


content/en/docs/get-started/local.md
```diff
@@ -18,14 +18,15 @@ PlanetScale provides [weekly builds](https://github.com/planetscale/vitess-relea
...
+sudo yum localinstall https://dev.mysql.com/get/mysql57-community-release-el7-9.noarch.rpm
+sudo yum install mysql-community-server
 ```
 
 _Vitess supports MySQL 5.6+ and MariaDB 10.0+. We recommend MySQL 5.7 if your installation method provides a choice._
 
 ## Disable AppArmor
 
-We recommend that you uninstall or disable AppArmor. Some versions of MySQL come with default AppArmor configurations that the Vitess tools don't yet recognize. This causes various permission failures when Vitess initializes MySQL instances through the mysqlctl tool. This is an issue only in test environments. If AppArmor is necessary in production, you can configure the MySQL instances appropriately without using `mysqlctl`:
+We recommend that you uninstall or disable AppArmor since it may causes permission failures when Vitess initializes MySQL instances through the `mysqlctl` tool. This is an issue only in test environments. If AppArmor is necessary in production, you can configure the MySQL instances appropriately without using `mysqlctl`:
```
aquarapid: "may cause"

sougou:
Some thoughts and comments:

* A downside of this approach is that the storyline gets lost. If someone starts off with "Getting Started", they can get lucky and follow the full story. But someone directly jumping into horizontal resharding may find it confusing. I don't have a better alternative yet though.

* 401_teardown.sh doesn't clean up everything. Some of the cleanup is done by 206_clean_commerce.sh. We need to address that part.

* I like the mysql launch improvement. We should definitely do the same thing for vtctlclient.

morgo:
> * A downside of this approach is that the storyline gets lost. If someone starts off with "Getting Started", they can get lucky and follow the full story. But someone directly jumping into horizontal resharding may find it confusing. I don't have a better alternative yet though.

I did put an Info tag to suggest using vertical first, but there could be some troubleshooting, because it will get ignored based on this pattern.

> * 401_teardown.sh doesn't clean up everything. Some of the cleanup is done by 206_clean_commerce.sh. We need to address that part.

I can fix this in code. I will modify 401 to also do 206_'s work if required.

> * I like the mysql launch improvement. We should definitely do the same thing for vtctlclient.

content/en/docs/user-guides/vertical-split.md
```diff
@@ -0,0 +1,181 @@
+---
+title: Vertical Split
+weight: 5
+---
+
+{{< info >}}
+This guide follows on from [get started with a local deployment](../../get-started/local). It assumes that the `./101_initial_cluster.sh` script has been executed, and you have a running Vitess cluster.
+{{< /info >}}
+
+Due to a massive ingress of free-trade, single-origin yerba mate merchants to your website, hipsters are swarming to buy stuff from you. As more users flock to your website and app, the `customer` and `corder` tables start growing at an alarming rate. To keep up, you’ll want to separate those tables by moving `customer` and `corder` to their own keyspace. Since you only have as many products as there are types of yerba mate, you won’t need to shard the product table!
```
morgo: We should remove this phrase. It's hard to translate.

morgo:
@sougou 401_teardown can now start from any point. Do you have any issues if we merge this PR?

sougou (approved):

PR #311, opened 2019-10-22
Fix typo (becasue -> because)

mfoo:
Noticed this while reading the docs.


morgo (approved):
Thank you for your contribution!

PR #316, opened 2019-10-26
New build from source guide guides (split macOS and ubuntu/debian)

morgo:
These are new build-from-source guides. It shows the progress we've made on trimming dependencies and simplifying things.

Some further improvements to the user experience:

1. I think we should change the `BUILD_JAVA` and `BUILD_PYTHON` defaults, since we don't require these to build and contribute (the docker bootstrap will need to turn them back on for its use case).

2. etcd should be removed from bootstrap and just become a dependency like MySQL is. We can reliably get it from distros, and that way it's patched and kept up to date. We will then modify the examples to search for etcd in $PATH. See: https://github.com/vitessio/vitess/pull/5365

3. Consul should also be gated (default to off, turned on inside docker).

4. This will make `proto` the only dependency that bootstrap needs to install. I don't have an answer for this one yet. Since it's the only dependency required, it feels like we should do something... especially since the version is pinned to [3.6.1](https://github.com/protocolbuffers/protobuf/releases/tag/v3.6.1) (July 2018). Feedback welcome.

Edit: This requires https://github.com/vitessio/vitess/pull/5365 to merge for the macOS version to work.

### DIFF
--- a/content/en/docs/contributing/build-from-source.md
+++ b/content/en/docs/contributing/build-from-source.md
@@ -77,29 +77,30 @@ brew install go automake git curl wget mysql57 etcd
 
 ## Build Vitess
 
-1. Navigate to the directory where you want to download the Vitess source code and clone the Vitess GitHub repo. After doing so, navigate to the `src/vitess.io/vitess` directory.
+Navigate to the directory where you want to download the Vitess source code and clone the Vitess GitHub repo. After doing so, navigate to the `src/vitess.io/vitess` directory.
 
-    ```sh
-    cd $WORKSPACE
-    git clone https://github.com/vitessio/vitess.git \
-        src/vitess.io/vitess
-    cd src/vitess.io/vitess
-    ```
-2. Build Vitess using the commands below. Note that the `bootstrap.sh` script needs to download some dependencies. If your machine requires a proxy to access the Internet, you will need to set the usual environment variables (e.g. `http_proxy`, `https_proxy`, `no_proxy`).
+```sh
+cd $WORKSPACE
+git clone https://github.com/vitessio/vitess.git \
+    src/vitess.io/vitess
+cd src/vitess.io/vitess
+```
 
-    Run the boostrap.sh script:
+Build Vitess using the commands below. Note that the `bootstrap.sh` script needs to download some dependencies. If your machine requires a proxy to access the Internet, you will need to set the usual environment variables (e.g. `http_proxy`, `https_proxy`, `no_proxy`).
 
-    ```sh
-    BUILD_PYTHON=0 BUILD_JAVA=0 ./bootstrap.sh
-    ```
+Run the boostrap.sh script:
 
-    Build Vitess:
+```sh
+BUILD_PYTHON=0 BUILD_JAVA=0 ./bootstrap.sh
+```
 
-    ```sh
-    # Remaining commands to build Vitess
-    source ./dev.env
-    make build
-    ```
+Build Vitess:
+
+```sh
+# Remaining commands to build Vitess
+source ./dev.env
+make build
+```
 
 ## Common Build Issues
 
--- a/content/en/docs/contributing/build-from-source.md
+++ b/content/en/docs/contributing/build-from-source.md
@@ -9,76 +9,71 @@ If you run into issues or have questions, we recommend posting in our [Slack cha
 
 The following sections explain the process for manually building Vitess on Linux and macOS. If you are new to Vitess, it is recommended to start with the [local install](../../get-started/local) guide instead.
 
-### Install Dependencies
+## Install Dependencies
 
 Many of the Vitess developers use Ubuntu or macOS desktops. If you would like to extend this guide for `yum` based distributions, please [send us a pull request](https://github.com/vitessio/website).
 
-#### Ubuntu and Debian
+### Ubuntu and Debian
 
 In addition, Vitess requires the following software and libraries:
 
-1.  [Install Go 1.12+](http://golang.org/doc/install).
+#### Install Go 1.12+
 
+[Install Go 1.12+](http://golang.org/doc/install).
 The version included in your OS distribution may be older than this. You can check by running `go version`.
 
-2.  We recommend that you uninstall or disable AppArmor since it may cause permission failures when Vitess initializes MySQL instances through the `mysqlctl` tool. This is an issue only in test environments. If AppArmor is necessary in production, you can configure the MySQL instances appropriately without using `mysqlctl`:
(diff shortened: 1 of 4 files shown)


content/en/docs/contributing/build-from-source.md
```diff
@@ -9,106 +9,71 @@ If you run into issues or have questions, we recommend posting in our [Slack cha
...
+#### Install Dependencies
 
-    ```sh
-    export PATH="/usr/local/opt/mysql@5.7/bin:$PATH"
-    export PATH=/usr/local/go/bin:$PATH
-    export GOROOT=/usr/local/go
-    ```
+Run the following command:
 
-7.  For the Vitess hostname resolving functions to work correctly, a new entry has to be added into the /etc/hosts file with the current LAN IP address of the computer (preferably IPv4) and the current hostname, which you get by typing the 'hostname' command in the terminal.
```
morgo, line 105: Need to get some history on this behavior. I assume it is fixed now.

content/en/docs/contributing/build-from-source.md
```diff
@@ -120,143 +85,36 @@ _Vitess supports MySQL 5.6+ and MariaDB 10.0+. We recommend MySQL 5.7 if your in
...
-    # On macOS with Homebrew
-    export VT_MYSQL_ROOT=/usr/local/opt/mysql@5.7
-    ```
-
-    Note that the command indicates that the `mysql` executable should be found at `/usr/local/opt/mysql@5.7/bin/mysql`.
-
-4. Run `mysqld --version` and confirm that you are running MySQL 5.7.
-
-5. Build Vitess using the commands below. Note that the `bootstrap.sh` script needs to download some dependencies. If your machine requires a proxy to access the Internet, you will need to set the usual environment variables (e.g. `http_proxy`, `https_proxy`, `no_proxy`).
+2. Build Vitess using the commands below. Note that the `bootstrap.sh` script needs to download some dependencies. If your machine requires a proxy to access the Internet, you will need to set the usual environment variables (e.g. `http_proxy`, `https_proxy`, `no_proxy`).
```
morgo: If we gate all the dependencies, we can simplify this line.

content/en/docs/contributing/build-from-source.md
```diff
@@ -9,106 +9,71 @@ If you run into issues or have questions, we recommend posting in our [Slack cha
...
-    ```
+**Notes:**
+* Vitess currently has some tests written in Python, but this dependency can be avoided by running the tests in Docker (recommended).
+* We will be using etcd as the topology service. The `bootstrap.sh` script can also install Zookeeper or Consul for you, which requires additional dependencies.
 
-#### Mac OS
+#### Install Docker
 
-1.  [Install Homebrew](http://brew.sh/). If your `/usr/local` directory is not empty and you haven't yet used Homebrew, you need to run the following command:
+[Install Docker](https://docs.docker.com/install/). This is only required to run the Vitess testsuite. Should you decide to skip this step, you will still be able to compile and run Vitess.
```
morgo: Can the ubuntu/debian version install docker from OS repo? Should work to simplify.

content/en/docs/contributing/build-from-source.md
```diff
@@ -9,106 +9,71 @@ If you run into issues or have questions, we recommend posting in our [Slack cha
...
-    - git
-    - pkg-config
-    - bison
-    - curl
-    - unzip
+Install dependencies required to build and run Vitess:
 
-    These can be installed with the following apt-get command:
+```sh
+sudo apt-get install -y mysql-server mysql-client make unzip g++ etcd curl
```
morgo: Need to verify this dependency list against a minimal debian/ubuntu, and a gated bootstrap.sh.

content/en/docs/contributing/build-from-source.md
morgo: I am going to keep it separate. It takes more steps to setup the user as part of the docker group etc. Keep it simple for now, as it is optional.

content/en/docs/contributing/build-on-centos.md
```diff
@@ -0,0 +1,241 @@
...
+make docker_test flavor=mysql57
+```
+
+Running the full suite currently takes 2+ hours to complete.
+
+## Common Build Issues
+
+{{< info >}}
+If you run into issues or have questions, we recommend posting in our [Slack channel](https://vitess.slack.com), click the Slack icon in the top right to join. This is a very active community forum and a great place to interact with other users.
+{{< /info >}}
```
deepthi: We already have this at the top of the page. Do we really want to repeat it?

content/en/docs/contributing/build-on-macos.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Install Homebrew and Dependencies
+
+[Install Homebrew](http://brew.sh/). From here you should be able to install:
+
+```
+brew install go automake git curl wget mysql@5.7 etcd
+```
+
+Follow the step to add mysql@5.7 to your PATH:
```
deepthi: This can be rephrased: `Add mysql@5.7 to your PATH:`

content/en/docs/contributing/build-on-macos.md
```diff
@@ -0,0 +1,208 @@
...
+make docker_test flavor=mysql57
+```
+
+Running the full suite currently takes 2+ hours to complete.
+
+## Common Build Issues
+
+{{< info >}}
+If you run into issues or have questions, we recommend posting in our [Slack channel](https://vitess.slack.com), click the Slack icon in the top right to join. This is a very active community forum and a great place to interact with other users.
+{{< /info >}}
```
deepthi: Do we really want to repeat this? It is already at the top of the page.

content/en/docs/contributing/build-on-ubuntu.md
```diff
@@ -0,0 +1,274 @@
...
+
+### Full testsuite
+
+To run the testsuite in Docker:
+
+```
+make docker_test flavor=mysql57
+```
+
+Running the full suite currently takes 2+ hours to complete.
```
deepthi, line 178: We should add that tests can be run locally on Ubuntu via `make test` / `make unit_test` / `make site_test`.

deepthi (approved):
Very nice! A few minor questions / suggestions.

content/en/docs/contributing/build-on-ubuntu.md
morgo: I think for now we should only support running tests for new contributors via Docker. I would like to remove this restriction in the future after we remove the Python 2.7 requirement.

PR #328, opened 2019-10-31
Add ant to brew installs on Mac OS contributing guide

BenMusch:
This dependency is required to install zookeeper locally

### DIFF
--- a/content/en/docs/contributing/build-on-macos.md
+++ b/content/en/docs/contributing/build-on-macos.md
@@ -20,7 +20,7 @@ The following has been verified to work on __macOS Mojave__. If you are new to V
 [Install Homebrew](http://brew.sh/). From here you should be able to install:
 
 ```
-brew install go automake git curl wget mysql@5.7 etcd
+brew install go automake git curl wget mysql@5.7 etcd ant
 ```
 
 Add `mysql@5.7` to your `PATH`:


morgo:
Thank you for your contribution!

Just so we are on the same page, the tutorial should not require zookeeper, provided that the bootstrap is run without Java:
```
BUILD_PYTHON=0 BUILD_JAVA=0 ./bootstrap.sh
```

The local example is also etcd-based, which is why this choice was made. In future of course, the default for `BUILD_JAVA` should be `0` so this step can be skipped.

Unless I am missing something here?

BenMusch:
@morgo That's correct -- I was interpreting the doc more as "everything you need to run vitess locally on a mac" so when I hit the zookeeper roadblock I updated it here

Would there be a better place to document this?

morgo:
My thoughts were:
The get-started and contribution guides benefit from being etcd based, because it's so much lighter on dependencies / is a fellow CNCF project.

Then there should be a User Guide for Switching the TOPO to zookeeper, and maybe another one for consul. It's the last item here on the user-guide improvements list: https://github.com/vitessio/website/issues/263

I think it is important to have guides to show how to switch it, because in many cases users will already have their favorite TOPO. Asking them to switch to another is never a compelling argument. It's just a balancing act with initial friction/simplicity.

What do you think?

BenMusch:
Agree with that reasoning, thanks for the insight! Closing this PR as a result

PR #351, opened 2019-11-07
Move orchestrator content to own location

morgo:
Part of #263

Text is not changed. I didn't even realize this content was here the doc is so long.


content/en/docs/user-guides/integration-with-orchestrator.md
```diff
@@ -0,0 +1,45 @@
...
+
+[Orchestrator](https://github.com/github/orchestrator) is a tool for managing MySQL replication topologies, including automated failover. It can detect master failure and initiate a recovery in a matter of seconds.
+
+For the most part, Vitess is agnostic to the actions of Orchestrator, which operates below Vitess at the MySQL level. That means you can pretty much [set up Orchestrator](https://github.com/github/orchestrator/wiki/Orchestrator-Manual) in the normal way, with just a few additions as described below.
+
+For the [Kubernetes](../../get-started/kubernetes) example, we provide a sample script to launch Orchestrator for you with these settings applied.
+
+## Orchestrator configuration
+
+Orchestrator needs to know some things from the Vitess side, like the tablet aliases and whether semisync is enforced (with async fallback disabled). We pass this information by telling Orchestrator to execute certain queries that return local metadata from a non-replicated table, as seen in our sample [orchestrator.conf.json](https://github.com/vitessio/vitess/blob/master/docker/orchestrator/orchestrator.conf.json):
```
bnlandry: I think these parens are a bit unclear. We can probably omit this detail, since these are just examples.

content/en/docs/user-guides/integration-with-orchestrator.md
```diff
@@ -0,0 +1,45 @@
...
+
+``` json
+"DetectClusterAliasQuery": "SELECT value FROM _vt.local_metadata WHERE name='ClusterAlias'",
+"DetectInstanceAliasQuery": "SELECT value FROM _vt.local_metadata WHERE name='Alias'",
+"DetectPromotionRuleQuery": "SELECT value FROM _vt.local_metadata WHERE name='PromotionRule'",
+"DetectSemiSyncEnforcedQuery": "SELECT @@global.rpl_semi_sync_master_wait_no_slave AND @@global.rpl_semi_sync_master_timeout > 1000000",
+```
+
+There is also one thing that Vitess needs to know from Orchestrator, which is the identity of the master for each shard, if a failover occurs.
```
bnlandry: maybe rephrase: Vitess also needs to know the identity of the master for each shard. This is necessary in case of a failover.

How does Vitess learn this "from Orchestrator"? Perhaps explaining how the information is communicated or gathered would be clearer.

content/en/docs/user-guides/integration-with-orchestrator.md
```diff
@@ -0,0 +1,45 @@
...
+
+From our experience at YouTube, we believe that this signal is too critical for data integrity to rely on bottom-up detection such as asking each MySQL if it thinks it's the master. Instead, we rely on Orchestrator to be the source of truth, and expect it to send a top-down signal to Vitess.
```
bnlandry: I think we can omit this entire paragraph, since it's not really task-oriented.

content/en/docs/user-guides/integration-with-orchestrator.md
```diff
@@ -0,0 +1,45 @@
...
+
+This signal is sent by ensuring the Orchestrator server has access to vtctlclient, which it then uses to send an RPC to vtctld, informing Vitess of the change in mastership via the [`TabletExternallyReparented`](../../reference/vtctl/#tabletexternallyreparented) command.
```
bnlandry: The passive voice here is unclear. Maybe steps here would be better:
1. Ensure Orchestrator has access to vtctlclient
2. Configure Orchestrator to inform vitess of the change in mastership via... (with sample JSON)

bnlandry (approved):
There are some places this could be clarified, as noted.

PR #357, opened 2019-11-12
Improve production planning

### LINKED ISSUE #335: Better document minimum requirements for Vitess
Both in terms of minimum components, and the requirements for the components. Most logical place is https://vitess.io/docs/user-guides/production-planning/

### LINKED ISSUE #354: Suggest 1 vCPU/core per 1500 QPS
On https://vitess.io/docs/user-guides/production-planning/ it was suggested that we could quote 1 vCPU per 1500 QPS for `mysqld` and `vttablet` (each).

It will vary of course, but I think this is probably helpful for users in doing back of the napkin planning. It is also fairly harmless if not exactly accurate (versus some other tuning recommendations which can be a bit.. hmm.)

morgo:
Fixes #354
Fixes #335

### DIFF
--- a/content/en/docs/user-guides/production-planning.md
+++ b/content/en/docs/user-guides/production-planning.md
@@ -11,9 +11,9 @@ aliases: ['/docs/launching/production-planning/','/docs/launching/']
 A highly available Vitess cluster requires the following components:
 
 * 2 VTGate Servers
-* A redundant Topology Service (i.e. 3x etcd servers)
-* 3x MySQL Servers with semi-sync replication enabled
-* 3x VTTablet processes
+* A redundant Topology Service (i.e. 3 etcd servers)
+* 3 MySQL Servers with semi-sync replication enabled
+* 3 VTTablet processes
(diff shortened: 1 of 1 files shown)


content/en/docs/user-guides/production-planning.md
```diff
@@ -6,11 +6,25 @@ aliases: ['/docs/launching/production-planning/','/docs/launching/']
 
 ## Provisioning
 
+### Minimum Topology
+
+A highly available Vitess cluster requires the following components:
+
+* 2 VTGate Servers
+* A redundant Topology Service (i.e. 3x etcd servers)
+* 3x MySQL Servers with semi-sync replication enabled
```
deepthi: What do we mean by 3x here?

morgo: I've removed the `x`. It isn't there for VTGate servers, and makes sense without it.

deepthi (approved):
lgtm

PR #359, opened 2019-11-14
Remove how-to content from VSchema reference doc.

bnlandry:
This content appears elsewhere under the VSchema user guide.


morgo (approved):
LGTM

PR #368, opened 2019-11-28
interim helm update

morgo:
I have managed to get helm 3.0 to work, with some small changes I will add in a PR soon.

Kubernetes 1.16 looks like it might be a little bit of work, so I will add a note for now.

### DIFF
--- a/content/en/docs/get-started/kubernetes.md
+++ b/content/en/docs/get-started/kubernetes.md
@@ -8,7 +8,7 @@ aliases: ['/docs/tutorials/kubernetes/','/user-guide/sharding-kubernetes.html']
 This tutorial demonstrates how Vitess can be used with Minikube to deploy Vitess clusters.
 
 {{< warning >}}
-Kubernetes 1.16 is not yet supported. We are working on fixing this in [issue #5411](https://github.com/vitessio/vitess/issues/5411), and recommend that you use an older version of minikube (v1.2.0) until it is fixed.
+Kubernetes 1.16 is not yet supported. We are working on fixing this in [issue #5411](https://github.com/vitessio/vitess/issues/5411).
 {{< /warning >}}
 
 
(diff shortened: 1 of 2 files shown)


derekperkins:
I think this might be a little misleading because you can't install the etcd-operator with helm 3 yet. I spent 30 minutes at the Vitess station trying to get things working with a guy from Digital Ocean before that was the final stopping point.
https://github.com/helm/charts/issues/19008

morgo:
> I think this might be a little misleading because you can't install the etcd-operator with helm 3 yet. I spent 30 minutes at the Vitess station trying to get things working with a guy from Digital Ocean before that was the final stopping point.
> [helm/charts#19008](https://github.com/helm/charts/issues/19008)

Hmm.. I thought I got it to work with helm 3. I'll have to take another look.

derekperkins:
IIUC, the installation works, but it uses a helm 2 hook to create the CRDs themselves. If you already had the definitions themselves, the rest would have worked.

derekperkins:
What would be ideal for the getting started guide would be to use the Kubernetes apii server as the topo server. @carsonoid started on an implementation but didn't finish it. I think one of the biggest questions is how much it might bloat the binary to include the k8s client.
https://github.com/carsonoid/vitess/tree/kubernetestopo

carsonoid:
I was just thinking about the K8s topo server implementation today. I would love to get a chance to complete it.

morgo:
@derekperkins PTAL again

content/en/docs/get-started/kubernetes.md
```diff
@@ -5,39 +5,55 @@ featured: true
...
+example/rbac/create_role.sh
+kubectl create -f example/deployment.yaml
+```
 
-* Install the MySQL client. On Ubuntu: `apt-get install mysql-client`
-* Install vtctlclient
+4. Install the MySQL client locally. For example, on Ubuntu: `apt-get install mysql-client`
+5. Install vtctlclient locally:
     * Install go 1.12+
     * `go get vitess.io/vitess/go/cmd/vtctlclient`
```
derekperkins, line 45: Does this start to break with go modules? For the future, we should provide a download link to the latest versioned release.

derekperkins:
LGTM

derekperkins (approved):

content/en/docs/get-started/kubernetes.md
morgo: Still worked for me, but maybe not the smartest way to do things. I will change it once we publish the new (smaller) package format.

PR #377, opened 2020-01-02
Update build-on-macos.md

rafael:
We also need to make sure that mvn is installed as well. 

Signed-off-by: Rafael Chacon <rafael@slack-corp.com>

### DIFF
--- a/content/en/docs/contributing/build-on-macos.md
+++ b/content/en/docs/contributing/build-on-macos.md
@@ -65,9 +65,10 @@ brew cask install adoptopenjdk8
 brew cask info java
 ```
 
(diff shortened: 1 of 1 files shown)


content/en/docs/contributing/build-on-macos.md
```diff
@@ -65,9 +65,10 @@ brew cask install adoptopenjdk8
 brew cask info java
 ```
 
-You will also need to install `ant`:
+You will also need to install `ant` and `mvn`:
 ```
 brew install ant
+brew install mvn
```
morgo: Should this be:
```suggestion
brew install ant mvn
```

PR #379, opened 2020-01-03
Update build-on-macos.md

rafael:
We also need to make sure that mvn is installed as well. 

Signed-off-by: Rafael Chacon <rafael@slack-corp.com>


morgo (approved):
LGTM

PR #383, opened 2020-01-08
Vitess 5.0 install docs

morgo:
In Vitess 5.0, we will no longer need to set VTROOT.

I have modified the binary install instructions to install the download in `/usr/local/vitess` and then copy the examples directory for your usage locally. I think this will start to make more sense as we stream-line the examples and remove the other files... and then we encourage users to make their own edits.

Several of the common issues experienced are no longer relevant, because VTROOT is no longer required. vtdataroot will default to $PWD/vtdataroot.


deepthi (approved):
Looks good.
I would change this line in the PR description: `vtdataroot will default to $PWD.`
VTDATAROOT defaults to $PWD/vtdataroot.

PR #397, opened 2020-02-18
Create VTExplain user guide doc.

bnlandry:


content/en/docs/user-guides/vtexplain.md
```diff
@@ -0,0 +1,157 @@
+---
+title: VTExplain
+aliases: ['/docs/user-guides/vtexplain/']
```
morgo: Please remove the aliases line. It is new content, so we don't need a redirect.

morgo (approved):
Minor comment, otherwise LGTM

gedgar:
@bnlandry @morgo 

We should probably note that `vtexplain` explicitly requires a keyspace name before the actual guts of the vschema, otherwise you'll get an error. 

Eg, this vschema has a keyspace name and plays nicely with vtexplain:
```
{
    "ks1": {
        "sharded": false,
        "tables": {
            "t1": {}
        },
        "vindexes": {}
    }
}
```
This does not:
```
{
    "sharded": true,
...
...
}
```


And will fail with this error:
`ERROR: initVtgateExecutor: json: cannot unmarshal bool into Go value of type map[string]json.RawMessage`

bnlandry:
I haven't made the title specific to validating queries, because it seems to me that this workflow is more general than that, since the output of VTExplain also expresses how Vitess executes query, in addition to validating that the query works. I have a plan to write a doc specifically targeting validation, which would reference this. Let me know if you think I should revisit this decision.

morgo (approved):
LGTM

PR #411, opened 2020-03-04
Redo VStream Diagram as SVG

celestehorgan:
Redo the VStream diagram as an SVG.


morgo (approved):
LGTM. Thanks for the improvement!

PR #415, opened 2020-03-10
Vreplication docs: MoveTables and Resharding

### LINKED ISSUE #334: Rename vertical split to Move Tables
It just moves tables between keyspaces. We discussed and think this is a better name.

morgo:
Fixes #334 

I incorporated the actual workflow steps here instead of the shell scripts. I like it more because it is more descriptive of the usage and easier to customize. The shell scripts add a level of indirection which makes it harder.


content/en/docs/user-guides/table-migration.md
```diff
@@ -0,0 +1,187 @@
...
+|        4 |           4 | SKU-1002 |    30 |
+|        5 |           5 | SKU-1002 |    30 |
++----------+-------------+----------+-------+
+```
+
+Notice that we are using keyspace `commerce/0` to select data from our tables.
+
+## Planning a Table Migration
+
+In this scenario, we are going to split the `commerce` keyspace into `commerce` and `customer` keyspaces. The tables `Customer` and `COrder` will be migrated into the newly created keyspace, and the `Product` table will remain in the `commerce` keyspace. This operation is online, which means that it does not block either read or write operations to the tables, __except__ for a small window during the final cut-over.
```
rohit-nayak-ps: Maybe useful to give an idea of the size of the cutover window based on production workloads ...

content/en/docs/user-guides/vertical-split.md
```diff
@@ -3,6 +3,10 @@ title: Vertical Split
 weight: 6
 ---
 
+{{< warning >}}
+In Vitess 6, Vertical Split became obsolete with the introduction of Table Migration! It is recommended to skip this guide, and continue on with the [table migration user guide](../table-migration) instead.
```
sougou: `Table Migration` -> "Move Tables". Change the link accordingly.

content/en/docs/user-guides/horizontal-sharding.md
```diff
@@ -3,6 +3,10 @@ title: Horizontal Sharding
 weight: 7
 ---
 
+{{< warning >}}
+In Vitess 6, Horizontal Sharding became obsolete with the introduction of Table Resharding! It is recommended to skip this guide, and continue on with the [table resharding user guide](../table-resharding) instead.
```
sougou: This has become "Resharding". We'll need a new section for that replaces the current horizontal sharding.

sougou (commented):
I'll try this out when everything gets checked in and give more updates.

content/en/docs/user-guides/resharding.md
```diff
@@ -0,0 +1,335 @@
...
+      "auto_increment": {
+        "column": "order_id",
+        "sequence": "order_seq"
+      }
+    }
+  }
+}
+```
+
+Note that we have now marked the keyspace as sharded. Making this change will also change how Vitess treats this keyspace. Some complex queries that previously worked may not work anymore. This is a good time to conduct thorough testing to ensure that all the queries work. If any queries fail, you can temporarily revert the keyspace as unsharded. You can go back and forth until you have got all the queries working again.
```
rohit-nayak-ps: When you say conduct thorough testing: is this an offline check based on the parameters you mentioned above?

content/en/docs/user-guides/resharding.md
```diff
@@ -0,0 +1,335 @@
...
+vtctlclient -server localhost:15999 ApplyVSchema -vschema_file vschema_commerce_seq.json commerce
+vtctlclient -server localhost:15999 ApplySchema -sql-file create_customer_sharded.sql customer
+vtctlclient -server localhost:15999 ApplyVSchema -vschema_file vschema_customer_sharded.json customer
+```
+
+## Create new shards
+
+At this point, you have finalized your sharded VSchema and vetted all the queries to make sure they still work. Now, it’s time to reshard.
+
+The resharding process works by splitting existing shards into smaller shards. This type of resharding is the most appropriate for Vitess. There are some use cases where you may want topin up a new shard and add new rows in the most recently created shard. This can be achieved in Vitess by splitting a shard in such a way that no rows end up in the ‘new’ shard. However, it's not natural for Vitess. We have to create the new target shards:
```
rohit-nayak-ps: Typo?: "you may want topin up" => "you may want to bring up" ?

content/en/docs/user-guides/resharding.md
```diff
@@ -0,0 +1,335 @@
...
+
+This left-justified approach allows you to have keyspace ids of arbitrary length. However, the most significant bits are the ones on the left.
+
+For example an `md5` hash produces 16 bytes. That can also be used as a keyspace id.
+
+A `varbinary` of arbitrary length can also be mapped as is to a keyspace id. This is what the `binary` vindex does.
+
+In the above case, we are essentially creating two shards: any keyspace id that does not have its leftmost bit set will go to `-80`. All others will go to `80-`.
+
+Applying the above change should result in the creation of six more vttablet instances.
```
rohit-nayak-ps: Might be helpful to expand on the six: one master, one replica, one rdonly tablet for each of the two shards

content/en/docs/user-guides/resharding.md
morgo: I've rewritten this a bit. There is a VTExplain doc that I linked to. In the event that someone doesn't follow it, I will also say "you can temporarily revert". This is a much more production-friendly way of saying it.

content/en/docs/user-guides/move-tables.md
```diff
@@ -0,0 +1,195 @@
+---
+title: MoveTables
+weight: 6
+---
+
+{{< info >}}
+This guide follows on from [get started with a local deployment](../../get-started/local). It assumes that the `./101_initial_cluster.sh` script has been executed, and you have a running Vitess cluster.
```
bnlandry: ...that you have a running Vitess cluster.

content/en/docs/user-guides/move-tables.md
```diff
@@ -0,0 +1,195 @@
...
+{{< /info >}}
+
+MoveTables is a new [VReplication](../../concepts/vreplication) workflow in Vitess 6, and obsoletes Vertical Split from earlier releases.
+
+This feature enables you to move a subset of tables between keyspaces without downtime. For example, after [initially deploying Vitess](../../get-started/local) your single commerce schema may grow so large that it needs to be split into multiple keyspaces.
```
bnlandry: after [Initially deploying Vitess(../../get-started/local), your...

content/en/docs/user-guides/move-tables.md
```diff
@@ -0,0 +1,195 @@
...
+
+As a stepping stone towards horizontally sharding (splitting a single table across multiple servers), it usually makes sense to split from having a single monolithic keyspace (`commerce`) to having multiple keyspaces (`commerce` and `customer`) that match your access pattern. For example, in our ecommerce system we know that `customer` and `corder` tables are closely related and growing at a high rate just by themselves.
```
bnlandry: Readers ignore parens. Try: "As a stepping stone towards **horizontally sharding**, or splitting a single table across multiple servers, it usually makes

content/en/docs/user-guides/move-tables.md
```diff
@@ -0,0 +1,195 @@
...
+
+Lets start by simulating this situation by loading sample data:
```
bnlandry: Overall style note: it's often more clear to avoid the first person plural, since it creates ambiguity about who is who. Also, it can be helpful to avoid using the past and present tense, since it causes the user to mentally jump around in the sequence, which can be disorienting in a how-to guide.

content/en/docs/user-guides/move-tables.md
```diff
@@ -0,0 +1,195 @@
...
++----------+-------------+----------+-------+
+```
+
+Notice that we are using keyspace `commerce/0` to select data from our tables.
+
+## Planning to Move Tables
+
+In this scenario, we are going to split the `commerce` keyspace into `commerce` and `customer` keyspaces. The tables `Customer` and `COrder` will be moved into the newly created keyspace, and the `Product` table will remain in the `commerce` keyspace. This operation is online, which means that it does not block either read or write operations to the tables, __except__ for a small window during the final cut-over.
+
+We could have equally decided to keep the `Customer` and `COrder` tables in the `commerce` keyspace, and created a new keyspace to move our `Product` table into. Which makes more sense often depends on the specifics of your environment. If the `Product` table is much larger, it will take more time to move, but in doing so you might also be able to migrate to newer hardware which has more headroom before you need to perform additional operations such as sharding. Similarly, if the `Customer` and `COrder` tables are updated at a more frequent rate, then this could also increase the move time.
```
bnlandry: I tend to think that how-tos should stick to a single-workflow. If we introduce alternative workflows, it can clutter the guide and cause confusion.

content/en/docs/user-guides/move-tables.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Impact to Production Traffic
+
+Another consideration in planning which tables to move is the modification rate. Internally a MoveTables operation is comprised of both a table copy and a subscription to all changes made to the table. Vitess uses batching to improve the performance of both table copying and applying subscription changes, but you should expect that tables with lighter modification rates to move faster.
```
bnlandry: Internally, a ...

content/en/docs/user-guides/move-tables.md
```diff
@@ -0,0 +1,195 @@
...
+
+During the active move process, data is copied from replicas instead of the master server. This helps ensure minimal production traffic impact.
```
bnlandry: Some details like this may be omitted, since they aren't required for the user to complete their task. It may be that we can split some of this information off into a concept doc that explains what happens during this operation.

content/en/docs/user-guides/move-tables.md
```diff
@@ -0,0 +1,195 @@
...
+
+During the `SwitchWrites` phase of the MoveTables operation, Vitess may be briefly unavailable. This unavailability is usually a few seconds, but will be higher in the event that your system has a high replication delay.
+
+## Create new tablets
+
+The first step in our MoveTables operation is to deploy new tablets for our `customer` keyspace. By convention, we are going to use the UIDs 200-202 as the `commerce` keyspace previously used `100-102`. Once the tablets have started, we can `-force` the first tablet to be the master:
```
bnlandry: Hard to read this shorthand: " we can `-force` the first tablet to be the master:"
Consider: "we can force the first tablet to be the master using the `-force` flag."

content/en/docs/user-guides/move-tables.md
```diff
@@ -0,0 +1,195 @@
...
+    MoveTables \
+    -workflow=commerce2customer \
+    commerce customer customer,corder
+
+sleep 2
+```
+
+## Phase 1: Switch Reads
+
+Once the MoveTables operation is complete, the first step in making the changes live is to _switch_ `SELECT` statements to read from the new keyspace. Other statements will continue to route to the `commerce` keyspace. By staging this as two operations, Vitess allows you to canary the changes and reduce the risks associated with changes. For example, you may have a different configuration of hardware or software on the new keyspace.
```
bnlandry: 'canary' as a verb may be unclear to some users. Consider "allows you to test the changes..."

content/en/docs/user-guides/move-tables.md
```diff
@@ -0,0 +1,195 @@
...
+# Works
+mysql --table < ../common/select_customer0_data.sql
+
+# Expected to Fail!
+mysql --table < ../common/select_commerce_data.sql
+```
+
+## Cleanup
+
+The final step is to remove the data from the original keyspace. As well as freeing space on the original tablets, this is an important step to eliminate potential future confusions. If you have a misconfiguration down the line and accidentally route queries for the  `customer` and `corder` tables to `commerce`, it is much better to return a table not found error, rather than return stale data:
```
bnlandry: ...it is much better to return a "table not found" error, rather than return stale data:

content/en/docs/user-guides/resharding.md
```diff
@@ -0,0 +1,333 @@
...
+
+{{< info >}}
+This guide follows on from [MoveTables](../../user-guides/move-tables) and [Get Started with a Local deployment](../../get-started/local). It assumes that several scripts have been executed, and you have a running Vitess cluster.
+{{< /info >}}
+
+## Preparation
+
+Resharding enables you to both _initially shard_ and reshard tables so that your keyspace is partitioned across several underlying [tablets](../../concept/tablet). A sharded keyspace has some additional restrictions on both [query syntax](../../reference/mysql-compatibility) and features such as `auto_increment`, so it is helpful to plan out a reshard operation diligently. However, you can always _reshard again_ if your sharding scheme turns out to be suboptimal.
+
+Using our example commerce and customer keyspaces, lets work through the two most common issues.
```
bnlandry, line 14: "let's"

content/en/docs/user-guides/resharding.md
```diff
@@ -0,0 +1,333 @@
...
+
+### Sequences
+
+The first issue to address is the fact that customer and corder have auto-increment columns. This scheme does not work well in a sharded setup. Instead, Vitess provides an equivalent feature through sequences.
+
+The sequence table is an unsharded single row table that Vitess can use to generate monotonically increasing ids. The syntax to generate an id is: `select next :n values from customer_seq`. The vttablet that exposes this table is capable of serving a very large number of such ids because values are cached and served out of memory. The cache value is configurable.
```
bnlandry: I think "monotonically increasing IDs" is easier to read.

content/en/docs/user-guides/move-tables.md
deepthi: you could also try i.e.

content/en/docs/user-guides/resharding.md
```diff
@@ -0,0 +1,333 @@
...
+
+The VSchema allows you to associate a column of a table with the sequence table. Once this is done, an insert on that table transparently fetches an id from the sequence table, fills in the value, and routes the row to the appropriate shard. This makes the construct backward compatible to how MySQL's `auto_increment` property works.
+
+Since sequences are unsharded tables, they will be stored in the commerce database. The schema:
```
bnlandry: Maybe: "Here is the schema:"

content/en/docs/user-guides/resharding.md
```diff
@@ -0,0 +1,333 @@
...
+INSERT INTO order_seq (id, next_id, cache) VALUES (0, 1000, 100);
+```
+
+Note the `vitess_sequence` comment in the create table statement. VTTablet will use this metadata to treat this table as a sequence.
+
+* `id` is always 0
+* `next_id` is set to `1000`: the value should be comfortably greater than the `auto_increment` max value used so far.
+* `cache` specifies the number of values to cache before vttablet updates `next_id`.
+
+Larger cache values perform better, but will exhaust the values quicker since during reparent operations the new master will start off at the `next_id` value.
```
bnlandry: exhaust the values quicker, since...

content/en/docs/user-guides/resharding.md
```diff
@@ -0,0 +1,333 @@
...
+      "type": "sequence"
+    },
+    "product": {}
+  }
+}
+```
+
+### Vindexes
+
+The next decision is about the sharding keys, aka Primary Vindexes. This is a complex decision that involves the following considerations:
```
bnlandry: sharding keys, or Primary Vindexes.

content/en/docs/user-guides/resharding.md
```diff
@@ -0,0 +1,333 @@
...
+
+* What are the highest QPS queries, and what are the where clauses for them?
```
bnlandry: Maybe: ...and what are the `WHERE` clauses for them?

bnlandry: Also throughout.

content/en/docs/user-guides/resharding.md
```diff
@@ -0,0 +1,333 @@
...
+
+* For the customer table, the most common where clause uses `customer_id`. So, it shall have a Primary Vindex.
+* Given that it has lots of users, its cardinality is also high.
+* For the corder table, we have a choice between `customer_id` and `order_id`. Given that our app joins `customer` with `corder` quite often on the `customer_id` column, it will be beneficial to choose `customer_id` as the Primary Vindex for the `corder` table as well.
+* Coincidentally, transactions also update `corder` tables with their corresponding `customer` rows. This further reinforces the decision to use `customer_id` as Primary Vindex.
+
+There are a couple of other considerations out of scope for now, but worth mentioning:
+
+* It may also be worth creating a secondary lookup Vindex on `corder.order_id`.
+* Sometimes the `customer_id` is really a `tenant_id` (i.e. your application is a SaaS, which serves tenants that themselves have customers). One key consideration here is that the sharding by the `tenant_id` can lead to unbalanced shards. You may also need to consider sharding by the tenant's `customer_id`.
```
bnlandry: Maybe: Sometimes the `customer_id` is really a `tenant_id`. For example, your application is a SaaS, which serves tenants that themselves have customers.

content/en/docs/user-guides/resharding.md
```diff
@@ -0,0 +1,333 @@
...
+
+Since the primary vindex columns are `BIGINT`, we choose `hash` as the primary vindex, which is a pseudo-random way of distributing rows into various shards. For other data types:
+
+* For `VARCHAR` columns, use `unicode_loose_md5`.
+* For `VARBINARY`, use `binary_md5`.
+* Vitess uses a plugin system to define vindexes. If none of the predefined vindexes suit your needs, you can develop your own custom vindex.
+
+## Apply VSchema
+
+Applying the new VSchema instructs Vitess that the keyspace is sharded, which may prevent some complex queries. If is a good idea to [validate this](../vtexplain) before proceeding with this step. If you do notice that certain queries start failing you can always revert temporaily by restoring the old VSchema. Make sure you fix all of the queries before proceeding to the Reshard process.
```
bnlandry: "It is a good idea to..."

bnlandry: If you do notice that certain queries start failing, you can always revert temporarily...

content/en/docs/user-guides/resharding.md
```diff
@@ -0,0 +1,333 @@
...
+```
+
+### Shard naming
+
+What is the meaning of `-80` and `80-`? The shard names have the following characteristics:
+
+* They represent a range, where the left number is included, but the right is not.
+* Their notation is hexadecimal.
+* They are left justified.
+* A `-` prefix means: anything less than the RHS value.
```
bnlandry: I would avoid "LHS" and "RHS". Maybe "left value" and "right value"

content/en/docs/user-guides/resharding.md
```diff
@@ -0,0 +1,333 @@
...
+ SHARD=80- CELL=zone1 KEYSPACE=customer TABLET_UID=$i ./scripts/vttablet-up.sh
+done
+
+vtctlclient -server localhost:15999 InitShardMaster -force customer/-80 zone1-300
+vtctlclient -server localhost:15999 InitShardMaster -force customer/80- zone1-400
+```
+
+### Shard naming
+
+What is the meaning of `-80` and `80-`? The shard names have the following characteristics:
```
bnlandry: I think that we could split this off into its own topic, so that this resharding guide could focus on resharding-specific tasks.

content/en/docs/user-guides/resharding.md
```diff
@@ -0,0 +1,333 @@
...
+# Examples 306_down_shard_0.sh
+
+for i in 200 201 202; do
+ CELL=zone1 TABLET_UID=$i ./scripts/vttablet-down.sh
+ CELL=zone1 TABLET_UID=$i ./scripts/mysqlctl-down.sh
+done
+
+```
+
+In this script, we just stopped all tablet instances for shard 0. This will cause all those vttablet and `mysqld` processes to be stopped. But the shard metadata is still present. We can clean that up with this command (after all vttablets have been brought down):
```
bnlandry: After Vitess brings down all vttablets, we can clean that up with this command:

sougou (approved):

PR #418, opened 2020-03-27
Improve replication guide

morgo:
- Rewrite to avoid the use of "slave" (use replica instead, but make sure it is correct usage, since replica is also a tablet type.. and there could be other rdonly tablets).

- Clarify that Vitess doesn't provide consistent reads on replicas (in the space of MySQL replication systems.. it is possible.)


content/en/docs/reference/mysql-replication.md
```diff
@@ -12,21 +12,19 @@ Vitess makes use of MySQL Replication for both high availability and to receive
...
+* The master will only accept writes if it has at least one replica connected, and configured correctly to send semi-sync ACKs. Vitess configures the semi-sync timeout to essentially an unlimited number so that it will never fallback to asyncronous replication. This is important to prevent split brain (or alternate futures) in case of a network partition. If we can verify all replicas have stopped replicating, we know the old master is not accepting writes, even if we are unable to contact the old master itself.
 
-   This is important to prevent split brain (or alternate futures) in case of a network partition. If we can verify all slaves have stopped replicating, we know the old master is not accepting writes, even if we are unable to contact the old master itself.
+* Tablets of type rdonly will not send semi-sync ACKs. This is intentional because rdonly tablets are not eligible to be promoted to master, so we want to avoid the case where a rdonly tablet is the single best candidate for election at the time of master failure. However, a split brain is possible when all rdonly tablets have transactions that none of replica tablets have.
```
deepthi: Can you rewrite the last sentence to make this a bit clearer? Maybe lead with something like: `If we allowed semi-sync ACKs from rdonly..`.
Otherwise, on a quick read it seems like a split brain is possible even when rdonly doesn't send semi-sync ACKs.

content/en/docs/reference/mysql-replication.md
```diff
@@ -12,21 +12,19 @@ Vitess makes use of MySQL Replication for both high availability and to receive
...
-
-These behaviors combine to give you the property that, in case of master failure, there is at least one other replica type slave that has every transaction that was ever reported to clients as having completed. You can then ([manually](../vtctl/#emergencyreparentshard), or with an automated tool like [Orchestrator](https://github.com/github/orchestrator)) pick the replica that is farthest ahead in GTID position and promote that to be the new master.
+These behaviors combine to give you the property that, in case of master failure, there is at least one other replica that has every transaction that was ever reported to clients as having completed. You can then ([manually](../vtctl/#emergencyreparentshard), or [integrate with Orchestrator](../..//user-guides/integration-with-orchestrator/) to pick the replica that is farthest ahead in GTID position and promote that to be the new master.
```
deepthi:
```suggestion
These behaviors combine to give you the property that, in case of master failure, there is at least one other replica that has every transaction that was ever reported to clients as having completed. You can then ([manually](../vtctl/#emergencyreparentshard), or [using Orchestrator](../..//user-guides/integration-with-orchestrator/) to pick the replica that is farthest ahead in GTID position and promote that to be the new master.
```

deepthi (approved):
A couple of suggestions. Looks good otherwise.

content/en/docs/reference/mysql-replication.md
morgo: I think it's not important, so I can just remove it.

PR #422, opened 2020-03-31
Improve compatibility overview

morgo:
This improves the compatibility overview page slightly:

- Remove top heading for consistency.
- Remove redundant "vitess supports" in headings.
- Change mysql link to MySQL.com
- Move sentence about Vtgate advertising MySQL to top level, since it applies to all.
- Add a sentence about MySQL 5.6 reaching end of life (not a good idea for new deployments, although Vitess will continue to provide support *for migrations in*). 


bnlandry (approved):
LGTM

PR #424, opened 2020-03-31
Add PDF embed shortcode

### LINKED ISSUE #421: Presentations page no longer embedding PDFs
On the [presentations page](https://vitess.io/docs/resources/presentations/), there used to be embedded PDFs.

I am not really sure when it stopped working (it might be a browser update on my side, or some of the recent CSS tweaks). I noticed it when working on #420

@celestehorgan @lucperkins do you mind investigating?

celestehorgan:
Closes #421. 

Note that [another solution was possible](https://discourse.gohugo.io/t/raw-html-getting-omitted-in-0-60-0/22032), but factoring out inline HTML whenever possible is the better long-term solution. I mention this just in case we run across it again in the site's source in future.

### DIFF
--- a/content/en/docs/resources/presentations.md
+++ b/content/en/docs/resources/presentations.md
@@ -8,19 +8,19 @@ aliases: ['/resources/presentations.html']
 
 Sugu Sougoumarane presents an overview of Vitess at Highload in Moscow.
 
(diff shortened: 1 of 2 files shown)


morgo:
Awesome! Thx for the quick turnaround.

morgo (approved):

PR #426, opened 2020-04-06
Add PlanetScale Vitess Operator

morgo:
It is a PlanetScale project.. not a Vitess project, but it provides a lot of value. We just need to point out that it is third party.

Changes made:
- Rename "Kubernetes" install method to helm in most places
- Include steps in resharding and move-tables.
- Change the heading tag of deployment-specific options to H3 so they show up in side menu (useful for now).
- Add new get-started guide.
- Remove vtctl api reference from menu (not strictly related, but as we add another install method, this is premium real estate. I will be refactoring this content in another PR.)

<strike>This will require a corresponding PR on the vitessio/vitess website, since there are some inconsistencies between the install methods. I am working on it.</strike> I reconciled it in such a way it works for all.

(I work for PlanetScale, but will seek reviewers outside of PlanetScale.)


sougou:
I think this is ok. There are people that want an operator and vitess doesn't have an answer.
In fact, I think we should develop an operator version of the local example. I may do this in my spare time.

content/en/docs/get-started/operator.md
```diff
@@ -0,0 +1,170 @@
...
+PlanetScale provides a [Kubernetes Operator for Vitess](https://github.com/planetscale/vitess-operator), released under the Apache 2.0 license. The following steps show how to get started using Minikube:
+
+## Prerequisites
+
+Before we get started, let’s get a few things out of the way:
+
+1. Install [Minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/) and start a Minikube engine:
+
+    ```bash
+    minikube start --cpus=8 --memory=11000 --disk-size=50g --kubernetes-version=v1.14.9
```
morgo: @sougou I am curious where the 11G memory requirement comes from? This will limit everyone but the very high end (16G laptops are still quite uncommon).

derekperkins (approved):
lgtm

PR #451, opened 2020-05-12
Update recent helm changes

morgo:
A few changes made, most fairly obvious from the diff:

- Reduced memory requirements, since we now use fewer pods (it now uses the default minikube of 3900MB, 2 vCPUs).
- Change the output to show the fewer pods
- Change to use pf.sh script
- Change the final delete step.
- Remove the vtctld script, since it's been removed from source.

I've also bumped the helm version, and tested on the latest minikube.

### DIFF
--- a/content/en/docs/get-started/kubernetes.md
+++ b/content/en/docs/get-started/kubernetes.md
@@ -14,7 +14,7 @@ Before we get started, let’s get a few things out of the way:
 1. Install [Minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/) and start a Minikube engine:
 
     ```bash
-    minikube start --cpus=4 --memory=4000
+    minikube start
     ```
 
     Note the additional resource requirements. In order to go through all the use cases, many vttablet and MySQL instances will be launched. These require more resources than the defaults used by Minikube.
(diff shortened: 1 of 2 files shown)


content/en/docs/get-started/kubernetes.md
```diff
@@ -96,27 +96,31 @@ You can check the state of your cluster with `kubectl get pods,jobs`. After a fe
...
-pod/zone1-commerce-0-init-shard-master-dh727   0/1     Completed   0          2m44s
-pod/zone1-commerce-0-replica-0                 5/6     Running     0          2m44s
-pod/zone1-commerce-0-replica-1                 5/6     Running     0          2m44s
-pod/zone1-commerce-0-replica-2                 5/6     Running     0          2m44s
+pod/commerce-apply-schema-initial-c9hrm        0/1     Completed   0          95s
+pod/commerce-apply-vschema-initial-tgk5h       0/1     Completed   0          95s
+pod/vtctld-58bd955948-56b77                    1/1     Running     0          95s
+pod/vtgate-zone1-c7444bbf6-25t7m               1/1     Running     3          95s
+pod/zone1-commerce-0-init-shard-master-7fzg9   0/1     Completed   0          95s
+pod/zone1-commerce-0-replica-0                 0/2     Running     0          95s
```
sougou: minor nit: these should be `2/2`

sougou (commented):
I'm thinking we should add two more changes:
* Add a VDiff step.
* Change MoveTables workflow to DropSources. Resharding doesn't need to be changed.

I'm ok to move forward with this if you want to do the above as a separate step.

morgo:
I've changed to DropSources. The examples used in CI are not using it yet - I will make this change now too.

(For vdiff: let's handle it separately).

sougou (approved):

PR #452, opened 2020-05-12
Update Community Call to CNCF links

morgo:
We currently use a Google Docs document hosted by PlanetScale, and a Zoom account hosted by Slack.

The CNCF has created accounts for both of these for our use.


deepthi (approved):
LGTM

PR #461, opened 2020-06-25
Remove clear references to slave

morgo:
There are still references to master (these will be harder to change due to an ambiguous/non clear replacement word).

There are also still references in `docs/reference/programs`, which will need to be updated separately since these are code-generated from vitess help output.


morgo:
This will be fixed in https://github.com/vitessio/website/pull/465

PR #476, opened 2020-07-21
Test PR Do Not Merge

zacharysarah:
🛑 TEST PR, DO NOT MERGE 🛑 

Testing netlify to see if it comments on successful preview deploy.

### DIFF
--- a/content/en/docs/overview/_index.md
+++ b/content/en/docs/overview/_index.md
@@ -1,7 +1,7 @@
 ---
 title: Overview
 description: High-level information about Vitess
-weight: 1
+weight: 2
 aliases: ['/docs/overview/scalingwithvitess/','/user-guide/introduction.html']
 ---
 


deepthi:
I love that we now get a PR comment with the netlify preview :100: 
I see that a netlify preview has also been added to the list of checks (at the end) and it leads to the correct project and preview.
The only outstanding issue is that we still have the old netlify check lingering (1 out of 7 checks). This means that only admins can merge PRs.
The problem is that that preview was linked to an old project, so we deleted the old project and this check is now orphaned. What we _should_ have done (but did not know) was to first delete all the webhooks from that project.
Can we file a ticket with netlify to get this cleaned up?

zacharysarah:
@deepthi That seems like the best approach if you don’t have access to the old project for whatever reason.

In the meantime, this PR has served its purpose, so let’s close it.

PR #486, opened 2020-08-03
updated location of recovery document

arindamnayak:


deepthi:
One more request. Can we list MySQL & ripple under pre-conditions?
Or right at the top we can put a "Supported databases" and list them there. It should be very clear to people that MariaDB is not supported, nor have we tested with any binlog server other than ripple.

arindamnayak:
> One more request. Can we list MySQL & ripple under pre-conditions?
> Or right at the top we can put a "Supported databases" and list them there. It should be very clear to people that MariaDB is not supported, nor have we tested with any binlog server other than ripple.

@deepthi , updated the document, please have a look. I havn't put the mysql and ripple together, Mentioned mysql at top and put ripple as pre-condition saying `currently it is tested with ripple` .

content/en/docs/reference/features/recovery.md
```diff
@@ -5,6 +5,9 @@ aliases: ['/docs/recovery/pitr','/docs/reference/pitr/']
 
 ## Point in time recovery
 
+### Support Databases
```
deepthi:
```suggestion
### Supported Databases
```

deepthi (changes requested):

content/en/docs/reference/features/recovery.md
arindamnayak: done

deepthi (approved):

PR #487, opened 2020-08-07
Add tabs

### LINKED ISSUE #440: UI toggle between helm and local installation features
These two guides are new to Vitess 6:
* https://vitess.io/docs/user-guides/move-tables/
* https://vitess.io/docs/user-guides/resharding/

Note that they are the same guide for a Helm installation as a local installation (which is a nice improvement over the previous guides, and makes maintenance easier).

However, in MoveTables there is 1 example specific to Helm and in Resharding there are 3 examples.

Is it possible to have a UI toggle so a user can see "helm" by default, but then switch to see "local"? This would make it easier on readers.

celestehorgan:
Closes https://github.com/vitessio/website/issues/440. 

This PR also:
-  bumps the version of the Bulma CSS framework, as I was testing something that ultimately didn't pan out.
- Uses external CDNs for JQuery dependencies, as these don't seem to load properly when referenced locally.


deepthi (commented):
This is great!
Could you move "Operator" to be first and "Helm" to be last in the tabs?

celestehorgan:
@deepthi will do!

deepthi (approved):
Thank you, this looks great!

PR #488, opened 2020-08-17
region_sharding: update tutorial to show resharding

deepthi:
- archive previous tutorial into 7.0 folder
- instructions to use the updated example in https://github.com/vitessio/vitess/pull/6565

Signed-off-by: deepthi <deepthi@planetscale.com>

### DIFF
--- a/content/en/docs/7.0/user-guides/region-sharding.md
+++ b/content/en/docs/7.0/user-guides/region-sharding.md
@@ -1,0 +1,294 @@
+---
+title: Region-based Sharding
+weight: 8
+---
+
+{{< info >}}
+This guide follows on from the Get Started guides. Please make sure that you have a [local](../../get-started/local) installation ready. You should also have already gone through the [MoveTables](../move-tables) and [Resharding](../resharding) tutorials.
+{{< /info >}}
+
+## Preparation
+Having gone through the Resharding tutorial, you should be familiar with [VSchema](../../concepts/vschema) and [Vindexes](../../reference/vindexes).
+In this tutorial, we will create a sharded keyspace using a location-based vindex. We will create 4 shards (-40, 40-80, 80-c0, c0-).
+The location will be denoted by a `country` column.
+
+## Schema
+We will create 2 tables in this example.
+```text
+CREATE TABLE customer (
+  id int NOT NULL,
+  fullname varbinary(256),
+  nationalid varbinary(256),
+  country varbinary(256),
+  primary key(id)
+  );
+CREATE TABLE customer_lookup (
+  id int NOT NULL,
+  keyspace_id varbinary(256),
+  primary key(id)
+  );
+```
+
+The customer table is the main table we want to shard using country. The lookup table will help us do that.
+
+## Region Vindex
+We will use a `region_json` vindex to compute the keyspace_id for a customer row using the (id, country) fields.
+Here's what the vindex definition looks like:
+```text
+    "region_vdx": {
+	"type": "region_json",
+	"params": {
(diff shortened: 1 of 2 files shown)


content/en/docs/user-guides/region-sharding.md
```diff
@@ -117,6 +63,13 @@ This is what the JSON file contains:
 ```
 The values for the countries have been chosen such that 2 countries fall into each shard.
 
+In this example, we are using 1 byte to represent a country code. You can use 1 or 2 bytes. With 2 bytes, 65536 distinct locations can be supported. The byte value of the country(or other location identifier) is prefixed to a hash value computed from the id to produce the keyspace_id.
+This will be primary vindex on the `customer` table. As such, it is sufficient for resharding, inserts and selects.
+However, we don't yet support updates and deletes using a multi-column vindex.
+In order for those to work, we need to create a lookup vindex that can used to find the correct rows by id.
```
jmoldow, line 69: Don't we also need the lookup vindex so that selects that don't include the `country` can be done efficiently?

content/en/docs/user-guides/region-sharding.md
```diff
@@ -257,6 +187,294 @@ mysql> show tables;
...
+      "type": "region_json",
+      "params": {
+        "region_bytes": "1",
+        "region_map": "/home/user/my-vitess/examples/region_sharding/countries.json"
+      }
+    }
+  },
+  "tables": {
+    "customer": {
+      "columnVindexes": [
```
jmoldow, line 310: Is this supposed to be `column_vindexes`? That is how it appears everywhere else in this tutorial.

content/en/docs/7.0/user-guides/region-sharding.md
```diff
@@ -0,0 +1,294 @@
+---
+title: Region-based Sharding
```
jmoldow, line 2: Should this go under `6.0`, since the new guide is for `7.0`?

jmoldow (commented):
As I suggested in the other PR:

I'm curious to see a discussion in the tutorial on the sharding strategy for the lookup table.

- Discussion about the latency implications of going cross-shard (and likely cross-region) in order to query the id lookup table.
- Tradeoffs of cross-replicating every lookup table shard into every region. What are the problems that come up here, with regards to consistency, in the face of replication lag.

sougou (approved):

content/en/docs/user-guides/region-sharding.md
deepthi: That's what I thought too, but this is what is returned by running the command.

content/en/docs/7.0/user-guides/region-sharding.md
deepthi: The example in the 7.0 release is still the old one. Updated example will only be available on master.

deepthi:
> As I suggested in the other PR:
> 
> I'm curious to see a discussion in the tutorial on the sharding strategy for the lookup table.
> 
> * Discussion about the latency implications of going cross-shard (and likely cross-region) in order to query the id lookup table.
> * Tradeoffs of cross-replicating every lookup table shard into every region. What are the problems that come up here, with regards to consistency, in the face of replication lag.

We are working on a separate page for lookup vindexes. TBH I still need to work out some of the answers, we'll make sure to include them on that page and update this page once I have them.

content/en/docs/user-guides/region-sharding.md
deepthi: Good point. I'll update.

PR #489, opened 2020-08-19
Update/clarify some parameter documentation.

aquarapid:
Not 100% accurate/complete yet, but better than before.


deepthi (approved):
Note: we should update vtgate program reference with the same content.

PR #496, opened 2020-08-26
Expand on the MoveTables user guide, giving better

aquarapid:
look at internals, and more complex use-cases.

### DIFF
--- a/content/en/docs/concepts/move-tables.md
+++ b/content/en/docs/concepts/move-tables.md
@@ -8,7 +8,7 @@ MoveTables is a new workflow based on VReplication. It enables you to relocate t
 
 It is recommended to keep tables that need to join on each other in the same keyspace, so typical candidates for a MoveTables operation are a set of tables which logically group together or are otherwise isolated.
 
-If you have multiple groups of tables as candidates, which makes the most sense to move may depend on the specifics of your environment. For example, a larger table will take more time to move, but in doing so you might be able utilitize additional or newer hardware which has more headroom before you need to perform additional operations such as sharding.
+If you have multiple groups of tables as candidates, which makes the most sense to move may depend on the specifics of your environment. For example, a larger table will take more time to move, but in doing so you might be able to utilitize additional or newer hardware which has more headroom before you need to perform additional operations such as sharding.
 
 Similarly, tables that are updated at a more frequent rate could increase the move time.
 
--- a/content/en/docs/concepts/move-tables.md
+++ b/content/en/docs/concepts/move-tables.md
@@ -2,13 +2,13 @@
 title: MoveTables
 ---
 
-MoveTables is a new workflow based on VReplication. It enables you to relocate tables between keyspaces without downtime.
+MoveTables is a new workflow based on VReplication. It enables you to relocate tables between keyspaces, and therefore physical MySQL instances, without downtime.
 
 ## Identifying Candidate Tables
 
-It is recommended to keep tables that need to join on eachother in the same keyspace, so typical candidates for a MoveTables operation are a set of tables which logically group together or are otherwise isolated.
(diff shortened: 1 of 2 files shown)


content/en/docs/concepts/move-tables.md
```diff
@@ -2,13 +2,13 @@
...
 
 ## Identifying Candidate Tables
 
-It is recommended to keep tables that need to join on eachother in the same keyspace, so typical candidates for a MoveTables operation are a set of tables which logically group together or are otherwise isolated.
+It is recommended to keep tables that need to join on each other in the same keyspace, so typical candidates for a MoveTables operation are a set of tables which logically group together or are otherwise isolated.
 
-If you have multiple groups of tables as candidates, which makes the most sense to move may depend on the specifics of your environment. For example, a larger table will take more time to move, but in doing so you might also be able to migrate to newer hardware which has more headroom before you need to perform additional operations such as sharding.
+If you have multiple groups of tables as candidates, which makes the most sense to move may depend on the specifics of your environment. For example, a larger table will take more time to move, but in doing so you might be able utilitize additional or newer hardware which has more headroom before you need to perform additional operations such as sharding.
```
rohit-nayak-ps: should be "you might be able ***to*** utilize additional or newer hardware"

content/en/docs/user-guides/move-tables.md
```diff
@@ -61,11 +61,26 @@ Notice that we are using keyspace `commerce/0` to select data from our tables.
 
 ## Planning to Move Tables
 
-In this scenario, we are going to split the `commerce` keyspace into `commerce` and `customer` keyspaces. The tables `Customer` and `COrder` will be moved into the newly created keyspace, and the `Product` table will remain in the `commerce` keyspace. This operation is online, which means that it does not block either read or write operations to the tables, __except__ for a small window during the final cut-over.
+In this scenario, we are going to `customer` keyspace to the `commerce` keyspace we already have.  This new keyspace will be backed by its own set of mysqld instances. We will then move the tables `customer` and `corder` from the `commerce` keyspace into the newly created `customer`. The `product` table will remain in the `commerce` keyspace. This operation happens online, which means that it does not block either read or write operations to the tables, __except__ for a small window during the final cut-over.
```
rohit-nayak-ps: should be: "we are going to ***add*** `customer` keyspace"

content/en/docs/user-guides/move-tables.md
```diff
@@ -133,32 +152,110 @@ for i in 200 201 202; do
...
+    }
+  ]
+}
+```
+
+Basically what the `MoveTables` operation has done is to create routing rules to explicitly route queries to the tables `customer` and `corder`, as well as the schema/table combos of `customer.customer` and `customer.corder` to the respective tables in the `commerce` keyspace.  This is done so that when `MoveTables` creates the new copy of the tables in the `customer` keyspace, there is no ambiguity about where to route requests for the `customer` and `corder` tables.  All requests for those tables will keep going to the original instance of those tables in `commerce` keyspace.  Any changes to the tables after the `MoveTables` is executed will be copied faithfully to the new copy of these tables in the `customer` keyspace.
+
+## Monitoring Progress (optional)
+
+In this example there are only a few rows in the tables, so the `MoveTables` operation only takes seconds. If the tables were large, you may need to monitor the progress of the operation.  There is simple way to get a percentage complete status, but you can estimate the progress by running the following against the master tablet of the target keyspace:
```
rohit-nayak-ps: should be: "There is ***no*** simple way to get a percentage complete status"

rohit-nayak-ps (changes requested):
Looks great.

3 minor corrections needed.

rohit-nayak-ps (approved):

PR #500, opened 2020-09-01
Add user guide for CreateLookupVindex

aquarapid:

### DIFF
--- a/content/en/docs/user-guides/backup-and-restore.md
+++ b/content/en/docs/user-guides/backup-and-restore.md
@@ -1,6 +1,6 @@
 	---
 title: Backup and Restore
-weight: 9
+weight: 10
 aliases: ['/user-guide/backup-and-restore.html']
 ---
 
--- a/content/en/docs/user-guides/createlookupvindex.md
+++ b/content/en/docs/user-guides/createlookupvindex.md
@@ -1,0 +1,477 @@
+---
+title: CreateLookupVindex
+weight: 9
+---
+
+{{< info >}}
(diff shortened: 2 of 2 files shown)


hallaroo:
### Comments
- This line is generating more questions for me about the differences in the consistent_lookup. I'd either remove it or link to an explanation if we have one: 'although that probably does not make much sense for a SKU column, where there might be many rows for the same SKU value.'
### Typos

- In the Blue info box it should read: Make sure _you_ are at the point where you have the sharded keyspace called customer setup.
- Change a that to an as: Note that as mentioned above, we do not have to tell Vitess about how to shard the actual backing table for the lookup Vindex or any schema to create _as_ it will do it automatically. 
- Adjusting a capitol letter and a period placement: (_In_ this case this table is empty, because the copy has finished _already.)_
- Missing apostrophe in a couple places: 'First _lets_ look at the streams to the first master tablet' and '_Lets_ look at this table:' and '_lets_ try a lookup on the sku column' and '_lets_ ensure that the lookup Vindex'
- Missing capitol letter: _That_ also scatters, as expected.

### Long sentences
- We should probably break up the second sentence: Internally, the CreateLookupVindex process uses VReplication for the backfill process, until the lookup Vindex is “in sync”_. After_ which the normal process for adding/deleting/updating rows in the lookup Vindex flows via the usual transactional flow when updating the “owner” table for the Vindex.
- Would also recommend breaking this up for readability: However, the VReplication status commands needs to operate on individual tablets. _Thus,_ we need to check which tablets we have in our environment:
- Again just breaking this up as it relates to actions to take: Note these commands will clean up all VReplication streams on these tablets. _You_ may want to filter by id if there are other streams running:

### Phrasing adjustments
- I would adjust 'Basically, exactly what we expected.' to the following: Basically, this should show exactly what we expected.

aquarapid:
This content was already merged in #501, but if there are any comments, please leave them here and I will fix them up.

sougou (approved):

PR #506, opened 2020-09-04
Add authorization UG;  overhaul user management UG, and split in two.

aquarapid:


hallaroo:
#### Rewording suggestions
- I would change: 'Note that the Vitess authorization via ACLs are applied at the VTTablet level, not VTGate, as authentication is.' to 'Note that the Vitess authorization via ACLs are applied at the VTTablet level, _as opposed to on VTGate (which is the location of Vitess' internal authentication)_.'
- Would add a bit more here: 'Used for testing ACL policies.' to '_This should be used for_ ...'
- Would add a 'This' here: '_This needs_ to be enabled for your ACLs to have any effect.' 

#### Typos
- Remove an extra emit: Set to true to check the table ACL at runtime, and _emit_ only emit the TableACLPseudoDenied metric if a request would have been blocked. 
- Missing an and (I think): This is the name of the ACL (aclname in the example above) _and_ is what needs to be specified in

content/en/docs/user-guides/authorization.md
```diff
@@ -0,0 +1,178 @@
...
+   enumerated or specified by regex.
+
+
+## VTTablet parameters for table ACLs
+
+Note that the Vitess authorization via ACLs are applied at the VTTablet
+level, as opposed to on VTGate, where authentication is enforced.
+There are a number of VTTablet commandline parameters that control the
+behavior of ACLs.  Let's review these:
+ * `-enforce-tableacl-config`:  Set this to true to ensure VTTablet will not
```
deepthi:
```suggestion
 * `-enforce-tableacl-config`:  Set this to `true` to ensure VTTablet will not
```

content/en/docs/user-guides/user-management.md
```diff
@@ -1,18 +1,21 @@
...
-Vitess uses its own mechanism for managing users and their permissions through VTGate. As a result, the `CREATE USER....` and
+Vitess uses its own mechanism for managing users and their permissions through
+ VTGate. As a result, the `CREATE USER....` and
 `GRANT...` statements will not work if sent through VTGate.
 
 ## Authentication
 
-The Vitess VTGate component takes care of authentication for requests so you will need to add any users that should have access
-to the Keyspaces via the command-line options to VTGate.
+The Vitess VTGate component takes care of authentication for requests so you
```
deepthi:
```suggestion
The Vitess VTGate component takes care of authentication for requests so we
```
(Since the rest of the doc uses `we`)

deepthi (commented):
Just a couple of nits.
Are you planning to add LDAP and TLS in a future PR?

aquarapid:
> Just a couple of nits.
> Are you planning to add LDAP and TLS in a future PR?

LDAP, yes.

TLS should go in it's own UG.

deepthi:
DCO needs to be fixed.

hallaroo (approved):
DCO passed

PR #517, opened 2020-09-14
contrib: clarify new coding standards

sougou:
Signed-off-by: Sugu Sougoumarane <ssougou@gmail.com>

### DIFF
--- a/content/en/docs/contributing/code-reviews.md
+++ b/content/en/docs/contributing/code-reviews.md
@@ -35,7 +35,7 @@ We use unit tests both to test the code and to describe it for other developers.
   * Attempt to cover every corner case the change introduces. 
   The thumb rule is: if it can happen in production, it must be covered.
 * Integration tests should ensure that the feature works end-to-end. 
-It must cover all the important use cases of the feature.
+They must cover all the important use cases of the feature.
 * A separate pull request into `vitessio/website` that updates the documentation is required if the feature changes or adds to existing behavior.
 
 ### Bug fixes
--- a/content/en/docs/contributing/code-reviews.md
+++ b/content/en/docs/contributing/code-reviews.md
@@ -2,30 +2,49 @@
 title: Coding Standards
 ---
 
-## Important rules
+## Backwards Compatibility
 
-Vitess is being used to power many mission-critical production workloads at very large scale. Moreover, many users deploy directly from the master branch. It is extremely important that the changes made by contributrors do not break any existing workloads.
+Vitess is being used to power many mission-critical production workloads at very large scale. 
+Moreover, many users deploy directly from the master branch. 
+It is very important the changes made by contributors do not break any existing workloads.
 
 In order to avoid disruption, the following concerns need to be kept in mind:
-* Does the change affect any external APIs? If so, make sure that the change satisfies the [compatibility rules](https://github.com/vitessio/enhancements/blob/master/veps/vep-3.md).
+* Does the change affect any external APIs? If so, make sure the change satisfies the [compatibility rules](https://github.com/vitessio/enhancements/blob/master/veps/vep-3.md).
 * Can the change introduce a performance regression? If so, it will be good to measure the impact using benchmarks.
 * If the change is substantial or is a breaking change, you must publish the proposal as an issue with a title like `RFC: Changing behavior of feature xxx`. Following this, sufficient time has to be given for others to give feedback. A breaking change must still satisfy the compatibility rules.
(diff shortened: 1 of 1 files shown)


content/en/docs/contributing/code-reviews.md
```diff
@@ -1,34 +1,49 @@
 ---
-title: Code Reviews
+title: Coding Standards
 ---
 
-Every GitHub pull request must go through a code review and get approved before it will be merged into the master branch.
+## Important rules
+
+Vitess is being used to power many mission-critical production workloads at very large scale. Moreover, many users deploy directly from the master branch. It is extremely important that the changes made by contributrors do not break any existing workloads.
```
shlomi-noach: typo and markdown
```suggestion
Vitess is being used to power many mission-critical production workloads at very large scale. Moreover, many users deploy directly from the `master` branch. It is extremely important that the changes made by contributors do not break any existing workloads.
```

content/en/docs/contributing/code-reviews.md
```diff
@@ -1,34 +1,49 @@
...
+
+In order to avoid disruption, the following concerns need to be kept in mind:
+* Does the change affect any external APIs? If so, make sure that the change satisfies the [compatibility rules](https://github.com/vitessio/enhancements/blob/master/veps/vep-3.md).
+* Can the change introduce a performance regression? If so, it will be good to measure the impact using benchmarks.
```
shlomi-noach, line 13: This is an issue I've been struggling with. What are the benchmarks? We unfortunately can't tell developers that they should check for performance regression if we don't have the framework to measure that.

content/en/docs/contributing/code-reviews.md
```diff
@@ -1,34 +1,49 @@
...
+* If the change is substantial or is a breaking change, you must publish the proposal as an issue with a title like `RFC: Changing behavior of feature xxx`. Following this, sufficient time has to be given for others to give feedback. A breaking change must still satisfy the compatibility rules.
```
shlomi-noach, line 14: Suggestion: formalize this even more by adding a `breaking change` label. Not sure if the casual user will have privileges to do that.

As alternative, we should use an issue template, which the developer must use, and which explicitly includes the question "is this a breaking change?"

content/en/docs/contributing/code-reviews.md
```diff
@@ -1,34 +1,49 @@
...
+* New features that affect existing behavior must be introduced "behind a flag". Users will then be encouraged to enable them, but will have the option to fallback to the old behavior if issues are found.
+* New features that do not affect existing behavior have more wiggle room. For example, you do not necessarily have to write tests during your initial commits. But all requirements must eventually be met before announcing the feature as ready.
 
 ## What to look for in a Review
 
-Both authors and reviewers need to answer these general questions:
+Every GitHub pull request must go through a code review and get approved before it will be merged into the master branch.
 
+Every pull request should meet the following requirements:
```
shlomi-noach, line 21: I'd like to further make a suggestion: pull requests should not come out of the blue. A PR should be associated with an issue. 
e.g. https://github.com/vitessio/vitess/issues/6661 is the tracking issue for the https://github.com/vitessio/vitess/pull/6668 PR. Reason is that an Issue is a better place to discuss the problem in general, the need for a change. A PR is a specific implementation detail. There can be multiple iterative PRs addressing the same issue. So, for example, my own https://github.com/vitessio/vitess/pull/6547 is _not_ a good example, because the discussion is entirely in the PR.

That's a slight overhead for th edeveloper (create an issue, create a PR), but I found that it is also a good preventive mechanism. The developer will open an issue, and I (as a maintainer) can tell them "this is not something that is within the interests of th eproject" or indicate some complexities in the problem, or indicate that there's actually a simple workaround. This may prevent the PR from taking place in the first place, saving time for the developer.

shlomi-noach (commented):
Some thoughts and comments inline

content/en/docs/contributing/code-reviews.md
```diff
@@ -1,34 +1,68 @@
...
+
+We use unit tests both to test the code and to describe it for other developers. 
+
+* Unit tests should:
+  * Demonstrate every use case the change covers.
+  * Involve all important units being added or changed.
+  * Attempt to cover every corner case the change introduces. 
+  The thumb rule is: if it can happen in production, it must be covered.
+* Integration tests should ensure that the feature works end-to-end. 
+It must cover all the important use cases of the feature.
```
deepthi:
```suggestion
They must cover all the important use cases of the feature.
```

systay (approved):

PR #521, opened 2020-09-21
Change image link URL for transport security model page

raunaqgupta:
Fix the image link reference URL to point to two folder down instead of one.


deepthi (approved):
thanks @raunaqgupta!

PR #533, opened 2020-09-29
removing outdated references

hallaroo:
Removing 'to be implemented' references from VReplication documenation


rohit-nayak-ps (approved):
lgtm

PR #555, opened 2020-10-06
reordering concepts nav

hallaroo:


deepthi:
What is the rationale for re-organizing? At first glance, it doesn't seem like concepts includes features/programs/components.

hallaroo:
I wanted to break the concepts up into more understandable chunks, rather than just having the long list so that it's easier to know where those concepts fit within Vitess. 

The categories were my initial try at making those chunks. Having the components in a clear components list would have helped me distinguish them when I was first starting to read about Vitess.

For the others I was trying to use the same breakdown as the one for the Reference pages (features and programs).

deepthi:
I don't think the same breakdown as `Reference` makes sense here. Having said that, I don't have an alternative proposal for breaking this down into sub-categories. 
I can see that some of the topics are both Programs and Components, but the `Program` part of it should be documented in `Reference` and not in this section.
Maybe @bnlandry can suggest a categorization that works better.

hallaroo:
This PR is too old but might be good to look at again in 2021

PR #559, opened 2020-10-11
Basic tracing UG / setup guide.

aquarapid:


hallaroo (approved):
Made changes and pushed them. @aquarapid - Let me know if you think it should be higher up on the list and I can change that and then merge it.

PR #573, opened 2020-10-22
Update region-sharding.md shell script 201 path

frouioui:
Modified the path for the shell script 201 that starts the resharding procedure.

The path used in Vitess's repository is the following one: `./examples/region-sharding/201_main_sharded.sh`


deepthi (approved):
Thanks for the fix!

frouioui:
No problem!

PR #580, opened 2020-10-27
fixing broken links

hallaroo:
Ran a broken link check against vitess.io. This is to fix the links reported as broken. 

Many of the broken links where in the Chinese documentation so I added a note to the best of my ability to indicate that the Chinese docs were no longer being maintained.


content/en/blog/2020-04-29-announcing-vitess-6.md
```diff
@@ -21,7 +21,7 @@ The Helm charts now default to using Kubernetes as the [Topology Service](https:
 This change also unlocked the adoption of Helm 3 and support for a wider range of Kubernetes versions, making installing Vitess much easier.
 
 ### General Availability of VReplication-based Workflows
-While VReplication made its appearance in Vitess 4, it has now been promoted from experimental to general availability and the documentation now points to [MoveTables](https://vitess.io/docs/user-guides/move-tables/) and [Resharding](https://vitess.io/docs/user-guides/resharding/).
+While VReplication made its appearance in Vitess 4, it has now been promoted from experimental to general availability and the documentation now points to [MoveTables](https://vitess.io/docs/user-guides/migration/move-tables/) and [Resharding](https://vitess.io/docs/user-guides/advanced-configuration/resharding/).
```
deepthi: This resharding link doesn't work (tried it in the preview).
Can we not use aliases so that we don't have to go back and edit all hyperlinks?

deepthi: @raunaqgupta what's the best practice?

deepthi: example: https://github.com/vitessio/website/blob/prod/content/en/docs/user-guides/_index.md
Each time the location of the file changed, an alias was added.

hallaroo: I'll update it (properly) this time and if @raunaqgupta confirms that aliases should be used I'll make aliases for all of the user guides I adjusted.

raunaqgupta: I'm sorry for the delayed response. The ONLY consideration is that any new version of links effectively make our existing google indexed links break or any external shares (on blog posts, etc). The best way to deal with this problem has been to define redirect rules. For instance on `docs.planetscale.com`, we did end up having a slightly different url scheme and ended up having these redirects:

image

I think aliases work exactly like 301 redirects? If so, definitely the way to go.

Update: Hugo aliases do work like that: https://gohugo.io/content-management/urls/#how-hugo-aliases-work

hallaroo: Awesome. I'll get the redirect rules set up. I think I saw the '_redirects' file for vitess.io in the repo already.

raunaqgupta: @hallaroo I think you wouldn't even need to touch that file. Just adding those aliases per file in the markdown frontmatter should do the trick.

hallaroo: oh ok - there are items in that '_redirects' files for user guides should I just ignore those or do I need to remove them too? Sorry for the n00b questions!

raunaqgupta: I havent' muddled with this part of the vitess docs either so new for me as well. The `_redirects` file is confusing. There are actually so many redirections:

For instance, this defines redirection points to the `schema-management` page

image

but the page itself has redirections of its own:

image

I think in this PR we should only edit the frontmatter of the files and not touch `_redirects` file.

hallaroo: Perfect. I just wanted to make sure I didn't mess anything up. Thank you!

raunaqgupta (approved):
LG

Issue #342, opened 2019-11-05
Support 2nd level menu item


morgo:
In https://github.com/vitessio/website/issues/331 (I won't get to it for a month, but planning now) the plan is to have an extra level of navigation:

```
Reference > Programs > vtctl
```

It looks like the navigation system currently only supports a 1st level of navigation.

doeg:
I can take a look at this one 😸

PR #583, opened 2020-11-01
Fix navigation issue on vitess.io docs for vreplication links

rohit-nayak-ps:
Added VReplication to hardcoded sections aliased to Reference. This fixes the problem where the navigation on the left was collapsing while selecting a VReplication link.

Also fixed a rendering issue where the navbar would flicker if logo loaded slowly.

### DIFF
--- a/content/en/docs/reference/compatibility/_index.md
+++ b/content/en/docs/reference/compatibility/_index.md
@@ -1,5 +1,6 @@
 ---
 title: Compatibility
 description: Reference documents for MySQL compatibility with Vitess 
+weight: 4
 ---
 
--- a/content/en/docs/reference/features/_index.md
+++ b/content/en/docs/reference/features/_index.md
@@ -1,5 +1,6 @@
 ---
 title: Features
 description: Reference documents for Vitess features
+weight: 1
 ---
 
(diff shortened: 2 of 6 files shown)


shlomi-noach (commented):
I'd really like to see a generic solution, other than `(eq $currentSection "VReplication"))`, or else this condition will blow up (already I have the same issue with "making schema changes" section collapsing)

rohit-nayak-ps:
> I'd really like to see a generic solution, other than `(eq $currentSection "VReplication"))`, or else this condition will blow up (already I have the same issue with "making schema changes" section collapsing)

Absolutely. CNCF told me it would take weeks for them to get to this issue and a fix might take a while because it was related to Hugo. Hence I looked at this again and was happy to get a quick fix since it looks bad right now. When we get the new site designed (I believe we are in the process of starting this with cncf) hopefully we can influence a better design.

shlomi-noach (approved):
Imprinting a mental note on your subconscious memory to remove the hard coded `(eq $currentSection "VReplication")` clause once CNCF generically solves the issue.

doeg:
> I'd really like to see a generic solution, other than (eq $currentSection "VReplication")), or else this condition will blow up (already I have the same issue with "making schema changes" section collapsing)


👋 I can take a look at this as part of https://github.com/vitessio/website/issues/342

PR #603, opened 2020-11-26
cp -r files in /usr/local/vitess (instead of mv)

jfg956:
When using mv, the ownership of files are preserved, which is not right.  By using cp -r, the files are copied (instead of moved).  As this is run as root (sudo), the files are now owned by root, and the current user cannot modified them.


deepthi:
We can approve and merge once the DCO is fixed.

jfg956:
> We can approve and merge once the DCO is fixed.

I fixed something, hopefully this is good for you.

deepthi (approved):
LGTM

PR #601, opened 2020-11-26
Add anchors to headings with a render hook

doeg:
**Update:** Seems like the Hugo version we're using in Netlify is a few months out of date + doesn't support render hooks; see https://github.com/vitessio/website/pull/601#issuecomment-734459182

--

Hugo offers a ["render hook templates" feature](https://gohugo.io/getting-started/configuration-markup#heading-link-example), which we can use to add anchors beside every heading. 

Here's what it looks like locally:

Screen Shot 2020-11-26 at 2 45 18 PM

Screen Shot 2020-11-26 at 2 49 32 PM

I briefly considered setting `visibility:hidden` unless the heading is hovered, since that looks a little cleaner, similar to what GitHub does when rendering Markdown. Opacity, however, is a better approach since it's more accessible and works on mobile.

PR #608, opened 2020-11-28
Upgrade to Hugo v0.78.2

doeg:
Upgrades to Hugo v.0.78.2. The [release notes](https://github.com/gohugoio/hugo/releases) do not mention any breaking changes. There are, however, tons of cool new features!

Building with Hugo v.0.78.2 works as expected on my machine. I will also check out the Netlify deploy once it's up. :) 

Upgrading Hugo will let us add section anchors by way of https://github.com/vitessio/website/pull/601, since render hooks were [added in v.0.71.0](https://github.com/gohugoio/hugo/releases/tag/v0.71.0).

We can likely add some makefile magic to verify that a user's local Hugo version matches the version in netlify.toml, but I'll leave that as a future enhancement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,9 @@ This repo houses the assets used to build the website at https://vitess.io.
 
 ## Running the site locally
 
-To run the website locally, you need to have the [Hugo](https://gohugo.io) static site generator installed (installation instructions [here](https://gohugo.io/getting-started/installing/)). Once Hugo is installed run the following:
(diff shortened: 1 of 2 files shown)


deepthi (approved):
I did a cursory check of the preview and it looks good. I'll approve, but @doeg can you leave a note here confirming that the preview looks good before merging?

Note: I checked a few other CNCF websites before approving, and they are on hugo 0.73+, so I see no objection to upgrading to the latest and greatest.

doeg:
@deepthi looks good to me!

PR #609, opened 2020-11-28
Show search input on all routes

doeg:
# Changes
- Shows the search input on all routes instead of just /docs, which closes #509
- Moves the search input to the right of the social icons, except on mobile where it appears first. This required some divergence from the [bulma.io](http://bulma.io/) styling, but not much. (Their breakpoints are quite nice!) 
- Fixes styling on the [DocSearch widget](https://docsearch.algolia.com/) for mobile (screenshot below)
- Updates search input placeholder text to be more specific about search scope ("Search the docs" instead of just "Search")
- Moves the search input's icon to the left, which looks cleaner when the "x" icon is shown

# Screenshots
| | Before (prod) | After (branch) |
|---|---|---|
| Landing page | searchbar-home-prod | searchbar-home-branch |
| /docs | searchbar-docs-prod | searchbar-docs-branch |
| /blog | searchbar-blog-prod | searchbar-blog-branch |
| Mobile | searchbar-mobile-prod | searchbar-mobile-branch |

PR #611, opened 2020-11-29
Fix mobile nav scroll behavior + logo size

doeg:
Two small fixes in this one!

First, #609 introduced an annoying little bug: when opening mobile nav, the content is scrolled to the bottom. This is because I was setting `flex-direction: column-reverse` on mobile to place the search input at the beginning (top) of the nav on mobile and leaving it and the end (right) on desktop. It's simpler to let the menu be `display: block` on mobile + set the flex direction on desktop (which already uses `display flex`)

| Prod | Branch | 
|----|----|
| mobile-nav-scroll-bug  | mobile-nav-scroll-fix | 

Second, the mobile navbar Vitess logo was getting both height and width set, which made it look a lil squashed. All we need is a height. 

| Prod | Branch | 
|----|----|
| vitess-logo-squishy | vitess-logo-regular |

### DIFF
--- a/assets/sass/style.sass
+++ b/assets/sass/style.sass
@@ -193,26 +193,25 @@ figure
   &.is-active
     position: fixed
 
-// Display the search input above other nav links for the mobile navbar,
-// plus fixes for the DocSearch widget to prevent clipping on mobile.
 +until($navbar-breakpoint)
-	.navbar-menu.is-active
-		display: flex
-		flex-direction: column-reverse
-
   #search-bar
     display: block
 
   .algolia-autocomplete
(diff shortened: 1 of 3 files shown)


doeg:
~~Ah, please hold off on reviewing this for now -- I just found a regression in the search widget on desktop. 🤔 Will update this PR when it's fixed.~~ This has been fixed :) 

Screen Shot 2020-11-29 at 10 50 55 AM

doeg:
Ok, fixed + redeployed to Netlify. Gotta love CSS!

Screen Shot 2020-11-29 at 10 55 31 AM

This is now ready for review again. 🙏 Thank you!

deepthi (approved):

PR #612, opened 2020-11-29
Add breadcrumb navigation to docs (for desktop)

doeg:
In reading the Hugo docs for https://github.com/vitessio/website/issues/342, I found an [example of breadcrumb navigation](https://gohugo.io/content-management/sections/#example-breadcrumb-navigation) which seems like a useful enhancement. There isn't an open GitHub issue for this, so I understand if you'd rather not merge it! :) 

A couple small details to note!

First, here's what it looks like on the /docs route where breadcrumbs are NOT shown. This is because /docs is the root of the breadcrumb navigation, so the "breadcrumbs" render as a single, unlinked `<li>` and it looked goofy.

Screen Shot 2020-11-29 at 1 01 29 PM

Here's what the breadcrumbs look like as you navigate through the doc hierarchy. As pointed out in #342, the "Reference" section of the docs has three levels of navigation, which still works just fine.

Screen Shot 2020-11-29 at 1 06 58 PM
Screen Shot 2020-11-29 at 1 07 00 PM
Screen Shot 2020-11-29 at 1 07 04 PM

Here's an example of a particularly long breadcrumb:

Screen Shot 2020-11-29 at 1 11 26 PM

Finally, I decided to hide the breadcrumbs on mobile/tablet. This isn't _ideal_ since I think they'd be particularly useful on mobile. However, [bulma.io's breadcrumb component](https://bulma.io/documentation/components/breadcrumb/) doesn't handle reflows very well: 

Screen Shot 2020-11-29 at 1 01 57 PM

 I spent a bit of time messing with the CSS, but didn't end up with anything I was happy with. I think mobile breadcrumbs deserve special consideration (like a dropdown, perhaps combined with the current orange dropdown?) and a separate PR. :)

### DIFF
--- a/assets/sass/style.sass
+++ b/assets/sass/style.sass
@@ -53,7 +53,7 @@ figure
 .has-extra-padding
   padding: 2.5rem
 
-.docs-article ul li
+.docs-article .content ul li
   margin-left: -.5rem
   list-style-type: square
 
--- a/layouts/partials/docs/header.html
+++ b/layouts/partials/docs/header.html
@@ -3,6 +3,34 @@
 {{ $language     := .Language }}
 {{ $editUrl      := printf "https://github.com/vitessio/website/tree/prod/content/%s/%s" $language $path }}
 <header class="docs-header">
+  {{/* Hide breadcrumbs on mobile since bulma.io doesn't handle overflows nicely for long crumbs */}}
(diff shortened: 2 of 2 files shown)


deepthi (approved):
Nice work!

PR #614, opened 2020-11-30
Fix overflow on docs sidebar + remove table of contents from section pages 

doeg:
Two very tiny fixes in this one!

First, changes the docs sidebar to use `overflow: auto`, since `overflow: scroll` on the docs sidebar shows the scrollbar placeholders even when no overflow is happening. 

| Before (prod) | Branch (after) |
|--|--|
| docs-sidebar-overflow-scroll | docs-sidebar-overflow-auto |

Second, this removes the table of contents sidebar (on the right) from "section" page templates. The table of contents always (?) empty on these pages and, in any case, somewhat redundant given that the pages themselves are a table of contents. :) Example: https://deploy-preview-614--vitess.netlify.app/docs/reference/

| Before (prod) | Branch (after) |
|--|--|
| docs-section-with-toc | docs-section-no-toc |

### DIFF
--- a/assets/sass/docs.sass
+++ b/assets/sass/docs.sass
@@ -28,7 +28,7 @@ html, body
     background-color: $dark
     flex: 0 0 15%
     border-right: 1px solid lighten($dark, 10%)
-    overflow: scroll
+    overflow: auto
 
     .docs-menu
       width: 100%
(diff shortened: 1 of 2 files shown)


deepthi (approved):
LGTM

PR #617, opened 2020-11-30
Fix occlusion of headings + navbar when jumping to heading anchors

doeg:
# Changes

Up until now, clicking an anchor link to a heading results in the heading being occluded by the sticky navbar. This is true for links in the table of contents (on the right side) and the new links on the headings added in https://github.com/vitessio/website/pull/601. Here's what that looks like: https://deploy-preview-617--vitess.netlify.app/docs/reference/features/vreplication/

header-offset-bug

The fix involves some admittedly obscure CSS contortions. Here's a good write-up on it: https://css-tricks.com/hash-tag-links-padding/ And here's what it looks like on desktop + mobile:

header-offset-fix
header-offset-fix-mobile

# One small caveat...

There is a _tiny_ bit of collateral damage from this. Pages other than the /docs pages (like Blog + Community) use a different navbar. As a result, the margin ends up being a bit _too_ generous. Here's a comparison of prod vs. this branch. You can see that prod snugs up the header + heading a little more closely: https://deploy-preview-617--vitess.netlify.app/blog/2020-10-27-announcing-vitess-8

| Prod | Branch |
|---|---|
| header-offset-prod-blog | header-offset-fix-blog | 

I _think_ there's likely some Hugo magic we could do to target docs-specific headings for the extra margin. However, I think it'd be cleaner + more consistent overall to unify the docs + non-docs navbars to use the same partial. 

I defer to y'all as to whether this is a dealbreaker, of course!

### DIFF
--- a/assets/sass/style.sass
+++ b/assets/sass/style.sass
@@ -221,3 +221,17 @@ figure
 
   &:hover
     opacity: 1
+
+// When jumping to anchor links, we need to account for the height of the sticky navbar.
+// Otherwise, the anchored header will be occluded by the navbar. 
+// See https://css-tricks.com/hash-tag-links-padding/
+$navbar-anchor-offset: calc(#{$navbar-height} + 16px)
+
+h1, h2, h3, h4, h5, h6
+  &:before
+    display: block
+    content: ' '
+    margin-top: calc(-1 * #{$navbar-anchor-offset})
+    height: $navbar-anchor-offset
+    visibility: hidden
+    pointer-events: none


deepthi:
This is great! It has been bugging me for a while following links on the vtctl command reference page (which your PR incidentally fixes).
I don't think the extra margin on the other pages is too big a deal (especially since we didn't even have anchors until yesterday), but do you have a sense of what it would take to unify the navbars?

doeg:
> do you have a sense of what it would take to unify the navbars?

@deepthi I don't think it would be challenging! Only thing is that the PR might be a little hefty since it'll restructuring the sidebars too. I think there are a handful of good reasons to do it, though, like consistent/simpler navigation, showing the i18n dropdown everywhere, this bug fix, etc.

I'm happy to write up an issue tomorrow and assign it to myself, if you like! :)

deepthi:
> 
> I'm happy to write up an issue tomorrow and assign it to myself, if you like! :)

That sounds great! We can merge this PR in the meantime :+1:

deepthi (approved):

Issue #625, opened 2020-11-26
Add applicable Vitess version to manual pages / documentation


bolt-leandro:
#### Feature Description

Vitess is currently changing at a fast pace so some parts of the written documentation are not backwards compatible or only apply to specific versions of Vitess.
It would be good if when writing/updating documentation, author could attach the Vitess version it applies to. A simple example is the MySQL docs where there is version selector ([https://dev.mysql.com/doc/refman/8.0/en/](https://dev.mysql.com/doc/refman/8.0/en/)). 

#### Use Case(s)

For example, currently, Vitess 8.0 gh-ost integration has specialized syntax while newer versions have @@ddl_strategy. The 8.0 GA version and the 'Managed, Online Schema Changes' man page are out of sync. One would have to check man page version history to find out about deprecated specialized syntax . This would probably cause some confusion to unaware users.

deepthi:
I looked at kubernetes.io and it seems like there may be a standard way of doing docs versioning with Hugo.
@doeg any thoughts on this?

doeg:
It looks like there are (at least?) two approaches to versioned content with Hugo. My $0.02 is that approach the first (git tags + separate deploys) is _probably_ simpler to set up, though perhaps less elegant from a user's perspective. Ultimately I think it'd be worth timeboxing an investigation into both to see which works best in practice. I could see the devil being in the details, particularly in terms of search + navigation. 

**1. Maintain previous versions as git tags + deploy static shapshots.**

I _think_ this is what https://kubernetes.io is doing. Specifically:

- [Docsy](https://www.docsy.dev/docs/adding-content/versioning/) is used for the version dropdown in the navbar
- Older versions of kubernetes.io documentation are deployed as a "static snapshot", for example: https://v1-15.docs.kubernetes.io.

One thing to note here is that the snapshots will be deployed with whatever layout they had at the time. 

**2. Version documentation with folder structure.**

Another approach is to "archive" docs using folder structure. For example:

content/docs/
├─ latest
├─ v8.0
├─ v7.0

This write-up summarizes some of the nuances with this approach, especially with links: https://github.com/google/docsy/issues/114#issuecomment-516592637

This thread is also useful + mentions how versioning works with the search widget: https://discourse.gohugo.io/t/documentation-site-versioning/5898/7

doeg:
[...]

Issue #618, opened 2020-12-02
Unify navbar partials


doeg:
Currently, we have two navbar components. 

First, [partials/docs/navbar.html](https://github.com/vitessio/website/blob/prod/layouts/partials/docs/navbar.html), which looks like this:

<img width="1792" alt="Screen Shot 2020-12-01 at 7 07 25 PM" src="https://user-images.githubusercontent.com/855595/100811371-80c97080-3408-11eb-9fe7-f43a638b6932.png">

And then [partials/navbar.html](https://github.com/vitessio/website/blob/prod/layouts/partials/docs/navbar.html) the default navbar used everywhere else. Note that the default navbar has dropdowns for the "Docs" and "Blog" sections and also has a dropdown for language selection:

<img width="1792" alt="Screen Shot 2020-12-01 at 7 08 47 PM" src="https://user-images.githubusercontent.com/855595/100811464-b4a49600-3408-11eb-899e-6e83460aff24.png">
<img width="1792" alt="Screen Shot 2020-12-01 at 7 08 53 PM" src="https://user-images.githubusercontent.com/855595/100811470-ba01e080-3408-11eb-830f-7809b7c44913.png">

Unifying the navbar would have a few benefits:
- We'd only need to have one partial, instead of two.
- We can unify the dropdown behavior across the entire site.
- We'll have multi-language selection on all pages.
- It will simplify bugfixes like the one described in https://github.com/vitessio/website/pull/617

PR #630, opened 2020-12-06
Update docs navigation to handle hierarchies of any depth

### LINKED ISSUE #342: Support 2nd level menu item

doeg:
Staged at https://deploy-preview-630--vitess.netlify.app/docs/

❤️ I've added a handful of reviewers to this as I'd like to get a lot of eyes on this! I'm not beholden to any of this implementation (including the styling) so would love your honest feedback! ~~cc @morgo also~~

# Changes
- Updates docs sidebar to handle hierarchies of any depth, which closes https://github.com/vitessio/website/issues/342 and addresses. This implementation, being generic, also removes some inconvenient hardcoding mentioned in https://github.com/vitessio/website/pull/583#issuecomment-720469238.

- Moves `content/docs/vtctl.md` to `content/docs/vtctl/_index.md`, which restructures the `vtctl` docs as a section + adds another layer of nesting. 

- Defines a new `docs_nav_title` front-matter variable, which lets pages specify a (shorter) title for the navbar that may differ from the page title.

- Defines a new `docs_nav_disable_expand` front-matter variable, and adds it to the "Older Version Docs" to hide the caret, since it gets treated like a section even though it only contains PDFs. We can likely :hocho: this when we improve version switching per https://github.com/vitessio/website/issues/625.

- Note that the size of the logo in the sidebar changed a bit. This way's a little better since we hardcode the size. Eventually I'd like to remove it as part of #618. 

# Screenshots

| | Before | After |
|-----|-----|-----|
| /docs | docs-root-prod | docs-root-branch |
| Troubleshooting | troubleshoot-prod | troubleshoot-branch | 
| vtctl schemas | vtctl-schema-prod | vtctl-schema-branch |
| vtctl schemas but on a slightly smaller screen! | vtctl-schema-small-prod | vtctl-schema-small-branch |
| /zh/docs | Screen Shot 2020-12-07 at 8 31 20 AM | Screen Shot 2020-12-07 at 8 31 18 AM | 

# Non-changes
There are a few things out of scope that {will, would} be cool to do:

- Handle expanding/collapsing sections of the navbar with JavaScript, so you can explore the hierarchy without requiring a full page refresh. It'll feel a lot faster. 

- We can also add expand/collapse all behaviour with JavaScript. And we can even make the sidebar's width adjustable. Or add headings (instead of the table of contents on the right.) Or... all kinds of things. 😎 

- Add a filter to the sidebar (distinct from the navbar search) to filter by title, like [terraform.io](https://www.terraform.io/docs/providers/index.html) for example:

    Screen Shot 2020-12-06 at 8 33 25 PM

# Trade-offs
I spent QUITE a bit of time trying to implement this with [Hugo's menu system](https://gohugo.io/content-management/menus/). There were, however, some big drawbacks that I ran into (and this may be because Hugo is still new to me):

- Using `sectionPagesMenu = "main"` as described in ["Section Menu for Lazy Bloggers"](https://gohugo.io/templates/menu-templates/#section-menu-for-lazy-bloggers) only added `.Site.Menus.main` for the topmost level of navigation. 🤔 Not useful.

- [Defining menu structure with front-matter](https://gohugo.io/content-management/menus/#add-content-to-menus) seemed to require a change to EVERY markdown file's front-matter... 🤔 🤔 Maybe there is a way to do this with cascading front-matter changes, but I couldn't get that to work. And it's still less desirable than inferring sidebar structure from directory structure (in my opinion), especially since we still [get the benefit of page params](https://gohugo.io/templates/menu-templates/#using-params-in-menus). 

- Finally, I _think_ the menu system has to be defined per-language...? Which meant that all of the front-matter changes to add `menu` have to be done for the translated content, too. 🤔 🤔 🤔 

In any case. It _does_ seem like using the menu system is the recommended way and I'd be happy to switch to that if there are good solutions to the issues above!

### DIFF
--- a/assets/sass/docs.sass
+++ b/assets/sass/docs.sass
@@ -1,5 +1,4 @@
 $docs-sidebar-logo-width: 30%
-$docs-sidebar-margin: 1.75rem
 $docs-sidebar-header-margin-bottom: 2rem
 $tan: #f5993a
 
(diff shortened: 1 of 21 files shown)


shlomi-noach:
I love this! Thank you @doeg ; I'll dig around a bit more to see that everything makes sense.

BTW, @morgo is no longer working on Vitess; @askdba is the person to ping now.

shlomi-noach:
Have taken this to the Vitess maintainers group -- sorry for the delays! I think this is lovely.

derekperkins (approved):
Looks great to me. Some of the flyout menus you mentioned would definitely be useful, so I'm glad for this cleanup that will enable that future work.

shlomi-noach (approved):

rohit-nayak-ps (approved):
Looks great!

shlomi-noach:
@doeg really appreciate you jumping on this one!

PR #632, opened 2020-12-08
fixing pdfs and rearranging order of advanced configuration user guides

hallaroo:


content/en/docs/pdfs/_index.md
```diff
@@ -4,8 +4,11 @@ description: PDFs of vitess.io/docs at the time of previous versions
 weight: 25
 ---
 
-To view a PDF of the state of the Vitess Documentation for previous versions of Vitess please follow the links below:
+The Vitess Documentation currently posted details everything as of the [latest release](https://github.com/planetscale/vitess-releases/releases).
```
deepthi: Ooh.. I don't think we want to refer to the planetscale weekly releases here.
Instead I would say that it reflects the latest code.
(We sometimes lag behind in docs, but we are trying to keep up)

content/en/docs/user-guides/configuration-advanced/user-management.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: User Management and Authentication
-weight: 5
+weight: 1
```
deepthi, line 3: What is the reason for moving User Management up? Resharding is already somewhat buried now compared to when it lived in the sidebar.

deepthi (dismissed):

content/en/docs/user-guides/configuration-advanced/user-management.md
hallaroo: This was a request from Jacques to move the authorization ones to the top 3 (specifically having the 2 authentication ones as 1 and 2 and having authorization as 3) and I made sure to keep resharding as the 4th so that it stays toward the top.

aquarapid (approved):

aquarapid (approved):

PR #634, opened 2020-12-10
Update faqs

### LINKED ISSUE #173: FAQ: Common Scenarios
Relating to [the thread](https://github.com/vitessio/vitess/issues/2811) with @sabutai's question on Slack (referencing this [issue](https://github.com/vitessio/vitess/issues/2811)):

* Add info about the workaround

```
sabutai   [15 hours ago]
For others who may be interested: We've found a workaround for this: By using Eclipselink QuerySequence, we were able to define a query to Vitess Sequences tables directly. This not only prevented SELECT LAST_INSERT_ID() call, we were also able to reduce number of DB trips since the application could request a bunch of IDs from Vitess, let's say 1000, so it will make only one call per 1000 inserts.
```

action: put faq for common scenarios back online

### LINKED ISSUE #257: Reincorporate FAQ changes
Extract out new FAQ questions from https://github.com/vitessio/website/pull/188

hallaroo:
I'm in the process of getting the second round of FAQs into FreshDesk and wanted to update the vitess.io site to point to there so that there is a link to the most up to date information.

I also copied any current FAQs from vitess.io into FreshDesk.  

I'll also be adding any previously requested FAQ updates to those in FreshDesk as part of this change.


deepthi (changes requested):
The FAQ in freshdesk looks really cool!
However, I would like us to keep what we have under FAQ. We can still add a link to the freshdesk FAQ but make it explicit that it is provided by PlanetScale.
Something like "PlanetScale hosts an FAQ for Vitess"

deepthi:
cc @vitessio/maintainers

hallaroo:
> The FAQ in freshdesk looks really cool!

Yay! The organization features and the ability to easily/quickly add SEO to each Freshdesk support page are really nice as well.

> However, I would like us to keep what we have under FAQ. We can still add a link to the freshdesk FAQ but make it explicit that it is provided by PlanetScale.

Could we maybe have an FAQ similar to https://www.mysql.com/about/faq/ (eg pointers on where to find things) and have any longer or more specific content in FreshDesk? Otherwise we run the risk of having detailed content split between sites (which I'd like to avoid).

One of the questions can then have the specific call out referenced:

> Something like "PlanetScale hosts an FAQ for Vitess"

The other option of having an FAQ, where all of the questions are written out as headers in list format on the website, I think runs the risk of adding a lot of additional content that would be hard to keep organized on vitess.io. There are something like 98 published articles now and I have one more round of questions to add. 

What do you think?

shlomi-noach:
FreshDesk looks really nice! At some point we might consider using GitHub Discussions for Q&A

content/en/docs/faq/_index.md
```diff
@@ -2,7 +2,44 @@
...
+
+We always enjoy having new contributors to Vitess. Just be sure to read the information [here](https://vitess.io/docs/contributing/) to start.
+
+If you are already familiar with Vitess and you'd like information on how to file a Pull Request or submit an Issue request check out the following links:
+
+* [Pull Requests](https://vitess.io/docs/contributing/github-workflow/#sending-pull-requests)
+* [Issue](https://vitess.io/docs/contributing/github-workflow/#submitting-issues)
+
+## What are good videos to watch to get started learning about Vitess?
+
```
deepthi, line 38: Can we add a line here about the videos we have on the vitess website itself and link to https://vitess.io/docs/resources/presentations/?

deepthi (changes requested):

content/en/docs/faq/_index.md
```diff
@@ -2,7 +2,44 @@
 title: FAQ
 description: Frequently Asked Questions about Vitess
 weight: 8
-featured: true
+docs_nav_disable_expand: true
 aliases: ['/docs/user-guides/faq/']
 ---
 
+## Where can I ask questions about Vitess?
```
deepthi, line 9: This is really nice content :100:

deepthi (approved):

Issue #636, opened 2020-12-12
Improve docs navigation on mobile


doeg:
The docs navigation menu is super useful, but it looks like there's an overflow bug: 

<img width="1392" alt="Screen Shot 2020-12-12 at 12 27 06 PM" src="https://user-images.githubusercontent.com/855595/101990538-651a6180-3c75-11eb-8e15-412f65a5b9f9.png">

hallaroo:
Hi @doeg - After the [PR](https://github.com/vitessio/website/pull/637) does that also fix this issue?

PR #637, opened 2020-12-12
Unify navbar partials and remove docs/navbar.html

### LINKED ISSUE #618: Unify navbar partials

doeg:
Closes https://github.com/vitessio/website/issues/618

Staged at https://deploy-preview-637--vitess.netlify.app/

# Changes

This PR removes the `docs/navbar.html` partial and updates all callsites to use the `navbar.html` partial instead. This makes the UI more consistent + should be easier for us to maintain, too.

A couple auxiliary changes:

- Minor layout adjustments to the docs layout to account for the fixed navbar. 

- Since the Vitess logo is now shown on all pages in the navbar, I removed the logo from the docs sidebar. As a bonus, this gives us more vertical height for the new, nested navs. 

- I biased to the `navbar.html` partial's approach to navbar links: namely, don't have dropdowns in the navbar. This also removes the featured articles from the mobile header, as shown in the screenshot below. The mobile /docs pages have their own navigation component (which is a little buggy: see https://github.com/vitessio/website/issues/636). The blogs page doesn't have a navigation component, which we could consider adding.

- It's worth noting that the translation dropdown is now shown on all pages, including /docs. 

# Screenshots

|   | Before (prod)  | After (branch)  |
|---|---|---|
| Home | header-nav-home-prod | header-nav-home-local |
| Docs | header-nav-docs-prod | header-nav-docs-local |
| Blog (no change) | header-nav-blog-prod | header-nav-blog-local |
| Small screen | header-nav-small-prod | header-nav-small-local |
| Docs (mobile) | header-nav-mobile-prod | header-nav-mobile-local |


deepthi (approved):
Very cool!

PR #647, opened 2020-12-21
Do vschema_customer_sharded.json before create_customer_sharded.sql.

jfg956:
When doing create_customer_sharded.sql (removing auto-increments) before vschema_customer_sharded.json (applying VIndex), there is a short period of time where INSERTs in the customer and corder tables would fail if the customer_id or order_id is not set.  Doing vschema_customer_sharded.json before create_customer_sharded.sql remove this failure possibility.


jfg956:
Related: https://github.com/vitessio/vitess/pull/7210

rohit-nayak-ps (approved):
lgtm

PR #594, opened 2020-11-18
Do Not Merge: new external db setup user guides

hallaroo:
Also need https://github.com/vitessio/vitess/pull/7056 for this

PR #648, opened 2020-12-22
Adding in more recent mysqldump version

hallaroo:
This is still being validated so I will hold off on merging until it is (and is reviewed).

### DIFF
--- a/content/en/docs/user-guides/configuration-basic/exporting-data.md
+++ b/content/en/docs/user-guides/configuration-basic/exporting-data.md
@@ -36,6 +36,12 @@ For example to export the `commerce` keyspace using the `mysqldump` binary from 
 $ mysqldump  --lock-tables=off --set-gtid-purged=OFF --no-tablespaces commerce > commerce.sql
 ```
 
+To do the same, but from Percona 8.0:
+
+```sh
+$ mysqldump  --lock-tables=off --add-locks=false --set-gtid-purged=OFF --no-tablespaces --add-drop-table=false -h x.x.x.x -u db_username -p db_name 
+```
+
 **NOTE:** You will be limited by the Vitess row limits in the size of the
 tables that you can dump using this method.  The default Vitess row limit is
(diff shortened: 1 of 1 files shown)


rohit-nayak-ps:
@hallaroo, may I suggest you convert this and #594  to `draft`s since they are not yet ready. Otherwise they show up as Ready and Pending Review and it may get accidentally reviewed and merged. Thanks!

content/en/docs/user-guides/configuration-basic/exporting-data.md
```diff
@@ -36,6 +36,12 @@ For example to export the `commerce` keyspace using the `mysqldump` binary from
 $ mysqldump  --lock-tables=off --set-gtid-purged=OFF --no-tablespaces commerce > commerce.sql
 ```
 
+To do the same, but from Percona 8.0:
```
mattlord, line 39: We should change this to MySQL 8.0 (Percona 8.0 is identical in this context).

content/en/docs/user-guides/configuration-basic/exporting-data.md
```diff
@@ -36,6 +36,12 @@ For example to export the `commerce` keyspace using the `mysqldump` binary from
...
+
+```sh
+$ mysqldump  --lock-tables=off --add-locks=false --set-gtid-purged=OFF --no-tablespaces --add-drop-table=false -h x.x.x.x -u db_username -p db_name
```
mattlord, line 42: From the paragraph above, I'm guessing that this should be:
```
$ mysqldump  --skip-network-timeout --skip-lock-tables --skip-add-locks --set-gtid-purged=OFF --no-tablespaces commerce > commerce.sql 
```

But I don't know the context for this... @aquarapid do you remember?

hallaroo:
@mattlord verified what works for both 5.7 and 8.0. Which is the current example with --add-drop-table=false added. I will open a new PR to make that change.

PR #672, opened 2021-01-19
CALL proc is now supported

systay:
This should not be merged until https://github.com/vitessio/vitess/pull/7287 has been merged


shlomi-noach (approved):
Can you please clarify if multiple result sets are supported?

systay:
Turns out CALL is always using multi results, so it wasn't anything we could push to later. I'll add something.

PR #684, opened 2021-01-24
Articles: add prev/next navigation links and use full viewport

rohit-nayak-ps:
1. Add prev/next buttons for navigating between articles in a section

Currently to go to the next section you need to click explicitly on the next link in the left navbar. However some sections, like the user-guides, are written as a single "e-book" which should be read in sequence. So navigation buttons to traverse though them should help.

2.  Articles use full viewport
 
At the moment articles only use 80% of the viewport, resulting in  white space on the right. It is unsure why this was done. I changed it to 100% to see what breaks, but am unable to find anything. So have not changed it to 100%.

nav1

nav2

### DIFF
--- a/layouts/partials/docs/header.html
+++ b/layouts/partials/docs/header.html
@@ -4,7 +4,7 @@
 {{ $editUrl      := printf "https://github.com/vitessio/website/tree/prod/content/%s/%s" $language $path }}
 <header class="docs-header">
   {{/* Hide breadcrumbs on mobile since bulma.io doesn't handle overflows nicely for long crumbs */}}
-  <nav class="breadcrumb is-hidden-touch" aria-label="breacrumbs">
+  <nav class="breadcrumb is-hidden-touch" aria-label="breadcrumbs">
     <ul>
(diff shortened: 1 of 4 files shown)


shlomi-noach (approved):
Nice!

doeg:
Ohh love this idea! Happy to review when you're done with the draft! 🎉

layouts/partials/docs/header.html
```diff
@@ -4,21 +4,21 @@
 {{ $editUrl      := printf "https://github.com/vitessio/website/tree/prod/content/%s/%s" $language $path }}
 <header class="docs-header">
   {{/* Hide breadcrumbs on mobile since bulma.io doesn't handle overflows nicely for long crumbs */}}
-  <nav class="breadcrumb is-hidden-touch" aria-label="breacrumbs">
+  <nav class="breadcrumb is-hidden-touch" aria-label="breadcrumbs">
```
doeg, line 7: Haha thanks 🙈

doeg (approved):
Great idea!

PR #692, opened 2021-01-28
adding PDFs to git repo

hallaroo:


content/en/docs/pdfs/_index.md
```diff
@@ -9,8 +9,8 @@ The Vitess Documentation currently posted should reflect the most recent code.
 
 To view a PDF of the state of the Vitess Documentation on the date of the general availability release for previous versions please follow the links below:
 
-- [Vitess Docs 9.0 on January 26, 2021](https://drive.google.com/file/d/18xPfVl-OpKX6QtjEUwFY_dU93Y-ZDPav/view?usp=sharing)
-- [Vitess Docs 8.0 on October 27, 2020](https://drive.google.com/file/d/1SMnw19uYYFl9vqYsbc67jIC-8VObfiwh/view?usp=sharing)
-- [Vitess Docs 7.0 on July 28, 2020](https://drive.google.com/file/d/1Uu5JU-tOzElRbLSnOS2ZaTE8BOI6DAWG/view?usp=sharing)
-- [Vitess Docs 6.0 on April 29, 2020](https://drive.google.com/file/d/1b3EwRxYc3uBfnCKDvMhLXWEZenvDYX2u/view?usp=sharing)
-- [Vitess Docs 5.0 on February 4, 2020](https://drive.google.com/file/d/1RaWnhR7C2RyaaXY6_82V8Kb03JjC5FCD/view?usp=sharing)
+- [Vitess Docs 9.0 on January 26, 2021](https://github.com/vitessio/website/blob/mp-pdf-folder/static/files/version-pdfs/Vitess-Docs-9.0-01-26-2021.pdf)
```
deepthi: This is referring to your branch. I believe you can link to relative paths for the files like how it is done here:
https://github.com/vitessio/website/blob/prod/content/en/docs/resources/presentations.md

deepthi (changes requested):
The paths to files need to be fixed.
Also, any reason we have the pdfs in the repo in two different folders? Can we have one copy?

content/en/docs/pdfs/_index.md
hallaroo: Yeah I only want one copy ... I thought I got them all from the same place but I must have copied and pasted incorrectly. I'll get that fixed ASAP.

hallaroo: This should now match the example you linked.

deepthi: Sorry if the example was misleading. We still want them to be just download links, we should not embed them using `ViewerJS`. Only the paths need to change (to be relative).

deepthi:
We still seem to have two copies of each file in this PR.
For example:
`static/files/version-pdfs/Vitess-Docs-5.0-02-04-2020.pdf` and `static/pdfs/Vitess Docs 5.0 (Feb 4 2020).pdf`

deepthi (approved):

PR #703, opened 2021-02-10
Add basic query consolidation doc

deepthi:


content/en/docs/user-guides/configuration-advanced/query-consolidation.md
```diff
@@ -0,0 +1,23 @@
...
+An identical query is one that is exactly the same including literals and bind variables.
+
+Flags:
+* `-enable_consolidator`: Defaults to true.
+* `-enable_consolidator_replica`: Only enable query consolidation on non-primary tablets.
+
+## Consistency
+
+Let us say user1 issues a read query, and user2 issues a write that changes the result that the first read query would get.
+Then user2 issues an identical read while user1's read is still executing. In this case the consolidator will kick in and user2 will get the result of user1's query thereby losing read-after-write consistency
```
jmoldow:
```suggestion
Then user2 issues an identical read while user1's read is still executing. In this case the consolidator will kick in and user2 will get the result of user1's query thereby losing read-after-write consistency.
```

deepthi: Thank you :)

content/en/docs/user-guides/configuration-advanced/query-consolidation.md
```diff
@@ -0,0 +1,23 @@
+---
+title: Query Consolidation
+weight: 16
+aliases: []
+---
+Query consolidation is a VTTablet feature.
```
hallaroo: Query consolidation is a VTTablet feature meant to protect your database from an overload caused by a spike in QPS for a specific query. 

Without this feature enabled such spikes can completely overwhelm the database. However, with this feature enabled the following will occur: when a vttablet receives a query, if an identical query is already in the process of being executed, the query will then wait.

**Note:** an identical query is one that is exactly the same, including literals and bind variables.

content/en/docs/user-guides/configuration-advanced/query-consolidation.md
```diff
@@ -0,0 +1,23 @@
...
+
+If there's a spike in QPS for a specific query it can completely overwhelm the database.
+The idea of query consolidation is to protect the database from such an overload. With this feature enabled, when vttablet receives a query, if an identical query is already in the process of being executed, 
+the query will wait. As soon as the first query returns from the underlying database, the result is sent to all callers that have been waiting.
+An identical query is one that is exactly the same including literals and bind variables.
+
+Flags:
+* `-enable_consolidator`: Defaults to true.
```
hallaroo, line 15: Flags:

* `-enable_consolidator`: Defaults to true.
* `-enable_consolidator_replica`: Only enable query consolidation on non-primary tablets.

hallaroo: This is just for formatting when I make the PDFs for older versions

content/en/docs/user-guides/configuration-advanced/query-consolidation.md
```diff
@@ -0,0 +1,23 @@
...
+* `-enable_consolidator_replica`: Only enable query consolidation on non-primary tablets.
+
+## Consistency
+
+Let us say user1 issues a read query, and user2 issues a write that changes the result that the first read query would get.
```
hallaroo: It is important to note that in some cases read-after-write consistency can be lost.

For example, if user1 issues a read query and user2 issues a write, that changes the result that the first read query would get, then user2 issues an identical read while user1's read is still executing.

In this case the consolidator will kick in and user2 will get the result of user1's query thereby losing read-after-write consistency.

content/en/docs/user-guides/configuration-advanced/query-consolidation.md
```diff
@@ -0,0 +1,23 @@
...
+Then user2 issues an identical read while user1's read is still executing. In this case the consolidator will kick in and user2 will get the result of user1's query thereby losing read-after-write consistency
+
+If the application is sensitive to this behavior then you can specify that consolidation should be disabled on the primary, i.e., `-enable_consolidator=false` and `-enable_consolidator_replica=true`
```
hallaroo: If the application is sensitive to this behavior, then you can specify that consolidation should be disabled on the primary using the following flags: `-enable_consolidator=false` and `-enable_consolidator_replica=true`.

hallaroo (changes requested):

content/en/docs/user-guides/configuration-advanced/query-consolidation.md
deepthi: I couldn't see the difference. What am I missing?

deepthi:
Thanks @hallaroo. I have accepted your suggestions except for the `Flags` section where I can't see the difference.

content/en/docs/user-guides/configuration-advanced/query-consolidation.md
deepthi: Added newline per chat with @hallaroo

hallaroo (approved):

PR #710, opened 2021-02-12
Update keyspaces.md

Anders-PlanetScale:
Signed-off-by: Anders Nelson <anders@planetscale.com>

Added -rename_tables option to the DropSources example in keyspaces.md to be consistent with available command options.


hallaroo (approved):
LGTM

PR #731, opened 2021-03-24
Code Generation Blog by Andres

askdba:
Signed-off-by: Alkin Tezuysal <alkin.tezuysal@gmail.com>


systay (approved):

shlomi-noach:
Opinion: don't label @vmg as "rockstar", as I think recently this tag can have a bit of negative association. Instead, use some proper English superlative/description.

PR #736, opened 2021-03-29
VDiff: Update parameter list to support recent enhancements to vdiff

rohit-nayak-ps:
The documentation for vdiff has been updated to add a few parameters that were added in the last few months but are not yet reflected in the documentation.


hallaroo (approved):
All formatting lgtm! And content was approved in previous diff if I recall correctly.

From @aquarapid: Approved for content

PR #754, opened 2021-05-01
Document supported where conditions in vreplication

kirs:
Similar to https://github.com/vitessio/vitess/pull/8014. Let's document all `WHERE` conditions that `vreplication` supports.


rohit-nayak-ps (approved):
ltgm

PR #762, opened 2021-05-14
Add a page about the new way we should handle release branches and bugfixes

systay:


content/en/docs/contributing/bugfixes.md
```diff
@@ -0,0 +1,45 @@
...
+## Support lifetime
+We support releases for 9 months, which translates to three versions.
+As an example, if we just released v10.0, we support the v9.x and v8.x versions.
+This means that when a bug is found, it needs to be fixed in all affected versions that are still supported.
+
+## Types of branches
+
+### Main 
+
+This is the main dev branch, and this is where we will cut the next release tag from. Feature branches should be merged into this branch.
```
askdba, line 14: Feature and Bugfix branches should be merged into this branch.

askdba (commented):
LGTM

content/en/docs/contributing/bugfixes.md
deepthi: If a bug is identified against a release and we decide to fix it there, it would go into that release branch. If we decide that it is not relevant to a release and is part of a new feature being developed on main, then it will be merged here.

content/en/docs/contributing/bugfixes.md
```diff
@@ -0,0 +1,45 @@
+---
+title: Bugfixes and Release Branches
+---
+
+## Support lifetime
+We support releases for 9 months, which translates to three versions.
+As an example, if we just released v10.0, we support the v9.x and v8.x versions.
+This means that when a bug is found, it needs to be fixed in all affected versions that are still supported.
```
deepthi, line 8: There is some ambiguity here. All bugs are not severe enough to be fixed on release branches. 
Specifically VEP-1 has this language:

Because the release cycle includes a new release every 12 weeks, most fixes will only be pushed to the master branch. An exception to this process will be made for high severity bugs, in which case a fix will be available in the weekly release for all currently supported versions.
High severity bugs in this context include CVEs, data corruption, wrong results or outage inducing issues. The Vitess team MAY decide to only backport bugs with higher user impact.

deepthi: So when a bug is found, we have to go through a few steps
- what is the release it is being reported against?
- what is the oldest supported release it exists in?
- is it severe enough to warrant a release branch fix?
- if yes, we fix it on the oldest release
- otherwise we fix it only on mainline

deepthi: We need to create a new VEP to amend the release process.
https://github.com/vitessio/enhancements/tree/master/veps
separately there is a request to extend the support window to 1 year.

jmoldow:
@systay @deepthi  Please complete and deploy this change? I keep failing at finding this information since it isn't on the website yet. Thank you!

PR #769, opened 2021-05-27
Update whatisvitess.md

ramicitrom:
Seems like a typo . . .


hallaroo (approved):
That does indeed seem to be a typo. This can be merged after the DCO check is fixed.

You'll need to add another commit following these steps: https://github.com/vitessio/website/pull/769/checks?check_run_id=2683472429

deepthi (changes requested):
This is not a typo. By deadlines we mean timeouts, so we should probably change it to that.

hallaroo:
I made the above change in https://github.com/vitessio/website/commit/f09f15d7d91c28d0e44b47d1a62d8f7c7e52f804 so closing this out

PR #780, opened 2021-06-30
arewefastyet blog post

frouioui:
## Description

Blog post to announce arewefastyet.

PR #787, opened 2021-07-16
Addition of small text mentioning benchmarks in the homepage

frouioui:
Adding a small text that mentions benchmarking in the website's home page as discussed in #780.


deepthi (approved):
Ideally we should highlight it more, but that almost requires a rewrite of the home page.
This is good for now.

frouioui:
Do you already have design ideas to make the home page more attractive @deepthi?

PR #791, opened 2021-07-29
OnlineDDL: SHOW VITESS_MIGRATION ... LOGS; removing premature description

shlomi-noach:
Documenting changes made in https://github.com/vitessio/vitess/pull/8493, which is available both in `v11` and `main`.

Removing premature description for OnlineDDL/VReplication: "Recoverable after failover." is not yet GA and not yet tested. We should not advertise it.


rohit-nayak-ps (approved):
lgtm

PR #794, opened 2021-08-02
PDF Add and reformat for v11

hallaroo:


content/en/docs/design-docs/query-serving/reserved-conn.md
```diff
@@ -25,15 +27,16 @@ If you are using an application or library that is issues these kind of `SET` st
...
 
 ### GET_LOCK() and reserved connections
-The MySQL locking functions allows users to work with user level locks. Since the locks are tied to the connection, and freeing lock has to be done in the same connection as the lock was acquired, use of these functions will force a connection to become a reserved connection. This connection is also kept alive so it does not time out due to inactivity. More information can be found [here.](/docs/design-docs/query-serving/locking-functions/).
 
+The MySQL locking functions allows users to work with user level locks. Since the locks are tied to the connection, and freeing lock has to be done in the same connection as the lock was acquired, use of these functions will force a connection to become a reserved connection. This connection is also kept alive so it does not time out due to inactivity. More information can be found [here.](/docs/design-docs/query-serving/locking-functions/).
```
deepthi:
```suggestion
The MySQL locking functions allow users to work with user level locks. Since the locks are tied to the connection, and the lock must be released in the same connection as it was acquired, use of these functions will force a connection to become a reserved connection. This connection is also kept alive so it does not time out due to inactivity. More information can be found [here.](/docs/design-docs/query-serving/locking-functions/).
```

deepthi (approved):
One suggestion on re-wording. The rest LGTM.

content/en/docs/design-docs/query-serving/reserved-conn.md
hallaroo: Changes made. Thank you!!!

PR #799, opened 2021-08-10
VReplication FAQ: initial set

rohit-nayak-ps:
A few commonly occurring issues that cause questions to support have been documented. This is a very small set to start with. Aim is to keep adding to these often by looking up Slack questions and based on feedback from others.

A new hugo layout shortcode `expand` has been added to support collapsible sections, which are ideal for a FAQ.


rohit-nayak-ps:
For reference, FAQ link: https://deploy-preview-799--vitess.netlify.app/docs/reference/vreplication/faq/

hallaroo (approved):
LGTM - I just indented one additional line as it was weird that only line 10 needed to scroll in that expanded FAQ section.

Also - do you think we can use this expand shortcode for the main FAQs https://vitess.io/docs/faq/?

shlomi-noach:
> Also - do you think we can use this expand shortcode for the main FAQs https://vitess.io/docs/faq/?

I think it makes more sense when:

- code snippets are involved, or:
- there's many questions.

I think https://vitess.io/docs/faq/ looks fine the way it is.

shlomi-noach:
@rohit-nayak-ps will be away for a while. I'm merging this PR on his behalf.

PR #800, opened 2021-08-10
Add Percona as supported DB

jonico:
* mention Percona support to be consistent with subsequent sections

### DIFF
--- a/content/en/docs/overview/whatisvitess.md
+++ b/content/en/docs/overview/whatisvitess.md
@@ -4,7 +4,7 @@ weight: 1
 featured: true
 ---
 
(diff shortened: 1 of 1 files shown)


deepthi:
@askdba WDYT? When we say MySQL we basically mean all variants (including Percona). Is it good to mention Percona explicitly?
If we do, perhaps we should make it clear that it is Percona MySQL server that is supported (since there are other DBs that Percona also provides, not to mention xtradb cluster etc.)

jonico:
@deepthi: I was just going through the docs section by section and in the [first section](https://vitess.io/docs/overview/whatisvitess/) only MySQL and Maria DB are mentioned whereas in the [third section](https://vitess.io/docs/overview/supported-databases/) Percona is mentioned as well.
Feel free to close if this is too much bike shedding 🤓

deepthi:
> @deepthi: I was just going through the docs section by section and in the [first section](https://vitess.io/docs/overview/whatisvitess/) only MySQL and Maria DB are mentioned whereas in the [third section](https://vitess.io/docs/overview/supported-databases/) Percona is mentioned as well.
> Feel free to close if this is too much bike shedding

That makes sense.

deepthi (approved):

PR #803, opened 2021-08-17
Multi-shard isolation user guide

hallaroo:
Adding in the user guide written by @aquarapid [here](https://github.com/aquarapid/vitess_examples/blob/master/shard_isolation_atomicity/shard_isolation_atomicity.md)


content/en/docs/user-guides/configuration-advanced/shard-isolation-atomicity.md
```diff
@@ -0,0 +1,186 @@
...
+
+This is typically not a big issue for most applications, since so-called read-after-write consistency is retained,  e.g.:
+
+  * if you performed a multi-value insert across multiple shards,
+  * **and** it completed successfully
+  * **then** if you issue a multi-shard (master) read after this, you should see the results of what you wrote across all shards (assuming nothing else deleted/updated those rows in the meanwhile)
+
+### Note  
+
+If you perform replica or rdonly reads instead of master reads (using the `@replica` or `@rdonly` Vitess dbname syntax extension), you will face the same issues you would if you read from a single MySQL replica instance. Accordingly, writes might not become visible for an extended period of time, depending on **replica lag**.  That being said, since Vitess helps you to keep your individual master instances smaller, replica lag should be less of an issue than it would be with a unsharded large MySQL setup.
```
deepthi:
```suggestion
If you perform replica or rdonly reads instead of master reads (using the `@replica` or `@rdonly` Vitess dbname syntax extension), you will face the same issues you would if you read from a single MySQL replica instance. Accordingly, writes might not become visible for an extended period of time, depending on **replica lag**.  That being said, since Vitess helps you to keep your individual master instances smaller, replica lag should be less of an issue than it would be with an unsharded large MySQL setup.
```

content/en/docs/user-guides/configuration-advanced/shard-isolation-atomicity.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Cross-shard atomicity
+
+When performing a write (`INSERT`, `UPDATE`, `DELETE`) across multiple shards, Vitess attempts to optimize performance, while also trying to ensure as much **atomicity** as possible. That is, Vitess will attempt to ensure that the whole write operation succeeds across all shards, or is rolled back.  However, if you think about what actually needs to happen across the multiple shards, achieving full atomicity across a (potentially large) number of shards can be very expensive. As a result, Vitess does not even try to guarantee cross-shard **isolation**, but rather focuses on trying to optimize cross-shard **atomicity**. The difference here is while the results of a single transaction might not become visible across all shards in the same instant, Vitess does try to ensure that write failures on a subset of the shards are:
```
deepthi:
```suggestion
When performing a write (`INSERT`, `UPDATE`, `DELETE`) across multiple shards, Vitess attempts to optimize performance, while also trying to ensure as much **atomicity** as possible. That is, Vitess will attempt to ensure that the whole write operation succeeds across all shards, or is rolled back.  However, if you think about what actually needs to happen across the multiple shards, achieving full atomicity across a (potentially large) number of shards can be very expensive. As a result, Vitess does not even try to guarantee cross-shard **isolation**, but rather focuses on trying to optimize cross-shard **atomicity**. The difference here is that while the results of a single transaction might not become visible across all shards in the same instant, Vitess does try to ensure that write failures on a subset of the shards are:
```

content/en/docs/user-guides/configuration-advanced/shard-isolation-atomicity.md
```diff
@@ -0,0 +1,186 @@
...
+
+As a result we might not even be able to tell the application with some certainty what happened. However, for some use-cases the performance of this option might be desirable. It is possible to select this behavior for individual DML statements in Vitess by using the special Vitess comment:
+
+```sh
+/*vt+ MULTI_SHARD_AUTOCOMMIT=1 */
+```
+
+It is not possible to make this the default behavior in Vitess; i.e. you will have to change your application code to take advantage of this option.
+
+In the same directory as this file, we have a script, `method1.sh`; which tries to use a sample vschema from `vschema.json` and SQL schema in `vschema.sql` to illustrate this method. Let's run this and inspect the output:
```
deepthi: It seems a bit abrupt that we move from theory to practice. It's still ok, but all the files referred here need to be included in the PR.

content/en/docs/user-guides/configuration-advanced/shard-isolation-atomicity.md
```diff
@@ -0,0 +1,186 @@
...
+multi-db transaction attempted: [target:{keyspace:"ks1" shard:"40-80" tablet_type:MASTER} transaction_id:1628034849705415307 tablet_alias:{cell:"explainCell" uid:2} target:{keyspace:"ks1" shard:"80-c0" tablet_type:MASTER} transaction_id:1628034849709028116 tablet_alias:{cell:"explainCell" uid:3} target:{keyspace:"ks1" shard:"-40" tablet_type:MASTER} transaction_id:1628034849700176113 tablet_alias:{cell:"explainCell" uid:1}]
+multi-db transaction attempted: [target:{keyspace:"ks1" shard:"40-80" tablet_type:MASTER} transaction_id:1628034849705415307 tablet_alias:{cell:"explainCell" uid:2} target:{keyspace:"ks1" shard:"80-c0" tablet_type:MASTER} transaction_id:1628034849709028116 tablet_alias:{cell:"explainCell" uid:3}]
+```
+
+As expected, we start getting errors since we are attempting a Vitess "transaction" across multiple shards.
+
+If we limited ourselves to writes that only target a single of the multiple shards, this would work fine, e.g.:
```
deepthi:
```suggestion
If we limited ourselves to writes that only target a single one of the multiple shards, this would work fine, e.g.:
```

deepthi (approved):
Excellent write-up. Thank you @aquarapid and @hallaroo.
A few minor changes are required, we can merge once those are done.

hallaroo:
I got the changes made but still need to add the referenced files to the vitess repo. I should be able to get that done tomorrow and then merge this.

askdba:
Should all `master` keywords be replaced with `primary` where applies in light of the inclusive naming project?

hallaroo:
> Should all `master` keywords be replaced with `primary` where applies in light of the inclusive naming project?

I just did a quick search to see what we have in other docs and we still have a number of references to master:

- https://vitess.io/docs/reference/programs/vttablet/
- https://vitess.io/docs/concepts/tablet/
- https://vitess.io/docs/troubleshoot/

So I'm not sure if we should also do a larger look across all the docs which this could be a part of? Doing a quick search via sublime for the keyword master I found 385 matches across 84 files. 

I'm happy to push this one and take an action item (that I could even start today) to do a different PR to do those replacements. What do you think @deepthi and @askdba ?

hallaroo:
And I just found @deepthi's draft PR 🤦🏼‍♀️ : https://github.com/vitessio/website/pull/805

deepthi (approved):

deepthi (approved):

PR #805, opened 2021-08-19
change master to primary in website docs

deepthi:
* Change `master` to `primary` in most places
* `multi-master` -> `active-active`
* Did not change old blog posts or `history.md`. We are not rewriting history here.
* Did not change `wip-docs`. We should delete most of them since that move was never completed, however there is some useful info there which should be brought back to the main docs.
* Did not change `zh` docs. These were made as a copy of the English docs for Chinese translation, but that effort has never been completed.
* Still TODO:
   -  We include `jquery-3.3.1.js` which uses `master` for a local variable.
   - `static/ViewerJS/webodf.js` has a bunch of occurrences of `master`, not sure how this can be fixed.


content/en/docs/reference/programs/vttablet.md
```diff
@@ -120,7 +120,6 @@ The following global options apply to `vttablet`:
 | -dba_idle_timeout | duration | Idle timeout for dba connections (default 1m0s) |
 | -dba_pool_size | int | Size of the connection pool for dba connections (default 20) |
 | -degraded_threshold | duration | replication lag after which a replica is considered degraded (only used in status UI) (default 30s) |
-| -demote_master_type | string | the tablet type a demoted master will transition to (default "REPLICA") |
```
hallaroo, line 123: Just confirming as I've mostly seen replacements. Is this just being fully removed? Or should there be a primary/leader based replacement flag?

hallaroo (approved):
I had just one question - once that's address lgtm 👍🏼

sougou (approved):

content/en/docs/reference/programs/vttablet.md
deepthi: The flag was deprecated several releases ago (probably in 8.0), so we went ahead and deleted it from the source code in a recent PR. This updates the docs to be consistent with that.

PR #813, opened 2021-08-24
adjusting spacing after word replacements

hallaroo:


mattlord:
This seems to have shifted things but they're still unaligned? 
- [Before](https://vitess.io/docs/user-guides/migration/move-tables/#show-our-old-and-new-tablets)
- [After](https://deploy-preview-813--vitess.netlify.app/docs/user-guides/migration/move-tables/#show-our-old-and-new-tablets)

mattlord (approved):

PR #818, opened 2021-08-31
updating outdated reference doc

hallaroo:
Updated as per https://github.com/vitessio/vitess/blob/e813ada00a6792624a2e3f05fc330184e71112e2/go/vt/vtgate/gateway.go#L40


Anders-PlanetScale (approved):
lgtm

PR #822, opened 2021-09-09
Addition of contributing reference for the parser and AST

GuptaManan100:
This PR adds a contributing reference for the parser and AST. It includes the files where the code lives, the make commands that help with the generated codes and the testing files for them.


frouioui (approved):
Looking good to me

PR #824, opened 2021-09-15
Adjusting v2 command docs

hallaroo:
Reorganizing to reduce duplicative content as the v2 commands had two unique .md pages with the same content.


Anders-PlanetScale (changes requested):
there are a few instances of "lags" I think it should just be lag

Select cells where replica lags are lower  should be:
Select cells where replica lag is lower

Reducing lags by pointing to PRIMARY should be:
reduce lag by pointing to PRIMARY

The rest looks good to me.

Anders-PlanetScale (approved):
Looks better, good catches on a few of these I missed.

PR #827, opened 2021-09-21
add information about schema tracking acl user

harshit-gangal:
Update the Schema tracking page with new flag required to be passed by user to make schema tracking work when ACL is enabled.

### DIFF
--- a/content/en/docs/reference/features/schema-tracking.md
+++ b/content/en/docs/reference/features/schema-tracking.md
@@ -18,6 +18,8 @@ The schema tracking functionality alleviates this issue and enable VTGate to pla
 ## VTGate
 
 Schema tracking is enabled in VTGate with the flag `-schema_change_signal`. When enabled, VTGate listens for schema changes from VTTablet.
+A change triggers a fetch query on vttablet on internal _vt schema.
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/features/schema-tracking.md
```diff
@@ -18,6 +18,8 @@ The schema tracking functionality alleviates this issue and enable VTGate to pla
 ## VTGate
 
 Schema tracking is enabled in VTGate with the flag `-schema_change_signal`. When enabled, VTGate listens for schema changes from VTTablet.
+A change triggers a fetch query on vttablet on internal _vt schema.
+If the Table ACL is enabled, then an exempted/allowed username needs to be passed to VTGate with flag `-schema_change_signal_user`.
```
frouioui, line 22: I think we should mention that `-schema_change_signal_user` defines the username used to send down queries to vttablet to request schema changes, while `-queryserver-config-acl-exempt-acl` is used to exempt certain users on vttablet's.

frouioui (approved):

content/en/docs/reference/features/schema-tracking.md
harshit-gangal: I do not want to add the acl flags as the users who enable it, already knows and there are 2 ways of doing it, so I refrain from mentioning it.

PR #847, opened 2021-10-19
region-sharding updates to switchtraffic

hallaroo:
adjusting from switch reads and writes to switchtraffic

### DIFF
--- a/content/en/docs/user-guides/configuration-advanced/region-sharding.md
+++ b/content/en/docs/user-guides/configuration-advanced/region-sharding.md
@@ -439,13 +439,16 @@ I0817 08:07:55.467790   15280 main.go:64] I0817 15:07:55.466993 reparent.go:274]
 
 ## Perform Resharding
 
-Once the tablets are up, we can go ahead with the resharding.
+Once the tablets are up, we can go ahead with the resharding:
+
 ```bash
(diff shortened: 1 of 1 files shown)


content/en/docs/user-guides/configuration-advanced/region-sharding.md
```diff
@@ -476,8 +479,7 @@ We have a running stream on tablet 200 (shard `-40`) that will keep it up-to-dat
 ## Cutover
 
 Once the copy process is complete, we can start cutting-over traffic.
-This is done in 2 steps, `SwitchReads` and `SwitchWrites`. Note that the commands are named for the tablet_types and not user operations.
-`Reads` is used for replica/rdonly, and `Writes` for primary. Read operations on primary will not be affected by a `SwitchReads`.
+This is done via [SwitchTraffic](../../../reference/vreplication/switchtraffic/). This replaced the previous SwitchReads and SwitchWrites commands with a single one. It is now possible to switch all traffic with just one command.
```
deepthi: Since we still have two steps in the example, we should call that out.
Something like:
"It is now possible to switch all traffic with just one command, however in this example we continue to do it in two steps because that is a more typical workflow."

deepthi (approved):

PR #877, opened 2021-11-12
Adding multi-cell topology information

FancyFane:
Adding additional clarification around configuring global and local topology services on different processes. This should assist users when creating multi-cell environments. 

Signed-off-by: FancyFane <fane@planetscale.com>

### DIFF
--- a/content/en/docs/reference/features/topology-service.md
+++ b/content/en/docs/reference/features/topology-service.md
@@ -18,8 +18,8 @@ The main contract for the Topology Service is to be very highly available and
 consistent. It is understood it will come at a higher latency cost and very low
 throughput.
 
-We never use the Topology Service as an RPC or queuing mechanism or as a storage 
-system for logs. We never depend on the Topology Service being responsive and 
+We never use the Topology Service as an RPC or queuing mechanism or as a storage
+system for logs. We never depend on the Topology Service being responsive and
 fast to serve every query.
 
 The Topology Service must also support a Watch interface, to signal when certain
@@ -32,7 +32,7 @@ We differentiate two instances of the Topology Service: the Global instance, and
 the per-cell Local instance:
 
 * The Global instance is used to store global data about the topology that
-  doesn’t change very often, e.g. information about Keyspaces and Shards. 
+  doesn’t change very often, e.g. information about Keyspaces and Shards.
   The data is independent of individual instances and cells, and needs
   to survive a cell going down entirely.
 * There is one Local instance per cell, that contains cell-specific information,
@@ -55,7 +55,7 @@ the tablets in that cell then need to be restarted so they re-initialize their
 topology records (but they won’t lose any MySQL data).
 
 If the Global Topology Service dies and is not recoverable, this is more of a
-problem. All the Keyspace / Shard objects have to be recreated or be restored. 
+problem. All the Keyspace / Shard objects have to be recreated or be restored.
 Then the cells should recover.
 
 ## Global data
@@ -180,7 +180,7 @@ new primary alias.
 
 Finding a tablet to serve the data is done in two stages:
 
-* vtgate maintains a health check connection to all possible tablets, and they 
+* vtgate maintains a health check connection to all possible tablets, and they
 report which Keyspace / Shard / Tablet type they serve.
 * vtgate also reads the SrvKeyspace object, to find out the shard map.
 
@@ -360,8 +360,8 @@ others wait for files with older ModRevisions to disappear.
 Leader elections also use a subdirectory, named after the election Name, and use
 a similar method as the locks, with ephemeral files.
 
-We store the proto3 binary data for each object (as the v3 API allows us to store 
-binary data).  Note that this means that if you want to interact with etcd using 
+We store the proto3 binary data for each object (as the v3 API allows us to store
+binary data).  Note that this means that if you want to interact with etcd using
 the `etcdctl` tool, you will have to tell it to use the v3 API, e.g.:
 
 ```
@@ -420,30 +420,51 @@ wait until the end of a polling cycle with that duration before returning.
 
 ## Running multi cell environments
 
-When running an environment with multiple cells, it's essential to first create
-and configure your global topology service, then register each local cell
-topology to the global topology. At a higher level overview:
+When running an environment with multiple cells, it is essential to first create
+and configure your global topology service. Then define each local topology
+service to the global topology. As mentioned perviously, the global
+and local topology service can reside on the same or sperate implementation of
+etcd, zookeeper, or consul. At a higher level overview:
 
-* Create the global keystore service using etcd, zookeeper, or consul. 
-* Use vtctld command to initialize it as the global topology service. NOTE: for
-  best practices the root dir should be set to `/vitess/global`.
-* For each cell create a local keystore service using a separate instance of
-  etcd, zookeeper, or consul.
-* For each cell register the new local topology service with the global topology
-  service using vtctl or vtctlclient commands. NOTE: for best practices the root
-  dir should be set to `/vitess/${CELL_NAME}`.
-* When starting vtgate/vttablet commands, you will provide the global topology
-  information, as well as the name of the cell. The information on the local
-  cell topology will be passed to the process from the global topology service
-  and your vtgate/vttablet will connect to the local topology service.
+* Create or locate an existing instance of etcd, zookeeper, or consul for the
+  global topology service.
+* Use vtctl client commands to initialize the global topology service,
+  providing the global topology implementation, and root directory.
+  NOTE: for best practices the root dir should be set to `/vitess/global`.
+* (Optional) For each cell create an instance of etcd, zookeeper, or consul for
+  the local topology service. This step is optional as you may use the existing
+  implementation used by the global topology service. If you create a new local
+  instances, the technologies must match. For example, if you are using etcd for
+  your global topology service then you must use etcd for your local topology
+  service.
+* For each cell, using the vtctl client commands, define the local topology
+  service with the global topology service. This is done by providing the global
+  topology service with the cell name, the local topology service, and the root
+  directory. NOTE: for best practices the local root dir should be set to
+  `/vitess/${CELL_NAME}` where `${CELL_NAME}` is the location of the cell
+  `us-east-1, eu-west-2, etc`.
+* When starting a vttablet instance you must provide the global topology service
+  as well as the `-tablet-path`, which implicitly includes the cell details.
+  With this information the vttablet process will read the local topology
+  details from the global topology server. The vttablet process is only
+  concerned with one local topology service.
+* When starting a vtgate instance, you will provide the global topology
+  service, as well as the `-cell` flag to explicitly provide the cell details.
+  With this information the vtgate process will retrieve the connection details
+  it needs to connect to the applicable local topology server(s). Unlike the
+  vttablet process, if you are watching more than one cell, in `-cells_to_watch`
+  you may connect to multiple local topology services.
+
 
 ### Simple local configuration
 
-For this example run through, we will be assuming a single node etcd service for
-the global and local topology services. NOTE: Production environments can and
-should be configured with multiple etcd nodes at the global and local levels.
+For this example run through, we will be using two etcd services one for
+the global and one for local topology service.
 
-1. Create the global etcd service to store our global values
+NOTE: Production environments can and should be configured with multiple
+instances at the global and local levels.
+
+1. Create the global etcd service
 
 ``` sh
 export VTDATAROOT="/vt"
@@ -459,7 +480,7 @@ etcd --enable-v2=true --data-dir ${VTDATAROOT}/etcd/global --listen-client-urls 
   ${OTHER_ETCD_FLAGS}
 ```
 
-2. Associate vtctld to the etcd service and initialize global topology values
+2. Configure vtctld to use the global topology service
 
 ``` sh
 vtctld -topo_implementation=etcd2 -topo_global_server_address=${GLOBAL_ETCD_SERVER} \
@@ -467,11 +488,11 @@ vtctld -topo_implementation=etcd2 -topo_global_server_address=${GLOBAL_ETCD_SERV
   ${OTHER_VTCTLD_FLAGS}
 ```
 
-3. Create a local etcd service to store our cell information
+3. Create a local etcd instance to store our cell information
 
 ``` sh
 CELL_NAME="US_EAST"
-CELL_TOKEN="SOMETHING_UNIQ_HERE_TOO"
+CELL_TOKEN="${CELL_NAME}"
 CELL_ETCD_IP="192.168.0.3"
 CELL_ETCD_SERVER="http://${CELL_ETCD_IP}:2379"
 CELL_ETCD_PEER_SERVER="http://${CELL_ETCD_IP}:2380"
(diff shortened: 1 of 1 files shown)


FancyFane:
This section could use additional clean up, and maybe even divide up the pages a bit more, but I'll mark this as "Ready for Review" for now, and we can create a new pull request for additional cleanup.

content/en/docs/reference/features/topology-service.md
```diff
@@ -418,7 +418,144 @@ use a long poll whose duration is set by the
...
 wait until the end of a polling cycle with that duration before returning.
 
-## Running in only one cell
+## Running multi cell environments
+
+When running an environment with multiple cells, it's essential to first create
+and configure your global topology service, then register each local cell
+topology to the global topology. At a higher level overview:
+
+* Create the global keystore service using etcd, zookeeper, or consul.
```
mattlord: We should be careful with terms as there's already enough to keep straight. We should use "global topology service" instead of "global keystone service".

content/en/docs/reference/features/topology-service.md
```diff
@@ -418,7 +418,144 @@ use a long poll whose duration is set by the
 `-topo_consul_watch_poll_duration` flag. Canceling a watch may have to
 wait until the end of a polling cycle with that duration before returning.
 
-## Running in only one cell
+## Running multi cell environments
+
+When running an environment with multiple cells, it's essential to first create
+and configure your global topology service, then register each local cell
+topology to the global topology. At a higher level overview:
```
mattlord: nit, but we're defining the local cell in the global topology (we define the name, the topo server, and root path for the cell).

content/en/docs/reference/features/topology-service.md
```diff
@@ -418,7 +418,144 @@ use a long poll whose duration is set by the
...
+
+* Create the global keystore service using etcd, zookeeper, or consul. 
+* Use vtctld command to initialize it as the global topology service. NOTE: for
```
mattlord: I would say `vtctl client command` here. `vtctl` is the stand-alone client and server combined, whereas `vtctlclient` is the client that works with a `vtctld` server. vtctld has a web UI (for humans) and a gRPC endpoint for clients (primarily vtctlclient).

mattlord: We often use vtctl as a short hand for vtctl/vtctlclient as they contain the same client code and commands, vtctlclient just needs to have a vtctld server provided for the server side.

content/en/docs/reference/features/topology-service.md
```diff
@@ -418,7 +418,144 @@ use a long poll whose duration is set by the
...
+  best practices the root dir should be set to `/vitess/global`.
+* For each cell create a local keystore service using a separate instance of
```
mattlord: Again,`s/keystore/topology/g`

content/en/docs/reference/features/topology-service.md
```diff
@@ -418,7 +418,144 @@ use a long poll whose duration is set by the
...
+  etcd, zookeeper, or consul.
+* For each cell register the new local topology service with the global topology
+  service using vtctl or vtctlclient commands. NOTE: for best practices the root
+  dir should be set to `/vitess/${CELL_NAME}`.
+* When starting vtgate/vttablet commands, you will provide the global topology
+  information, as well as the name of the cell. The information on the local
```
mattlord: To be clear, I would say the name of the cell where that component (vttablet, vtgate, etc) will live.

content/en/docs/reference/features/topology-service.md
```diff
@@ -418,7 +418,144 @@ use a long poll whose duration is set by the
...
+  cell topology will be passed to the process from the global topology service
```
mattlord: Nit, but I would say it will be read from the global topology service as the caller will request it via "cells_to_watch" (vtgate will often watch all of them whereas a vttablet only cares about its own cell -- which is why you must use a cell alias for multi-cell tablet driven operations like MoveTables).

content/en/docs/reference/features/topology-service.md
```diff
@@ -418,7 +418,144 @@ use a long poll whose duration is set by the
...
+GLOBAL_ETCD_PEER_SERVER="http://${GLOBAL_ETCD_IP}:2380"
+
+etcd --enable-v2=true --data-dir ${VTDATAROOT}/etcd/global --listen-client-urls ${GLOBAL_ETCD_SERVER} \
+  --name=global --advertise-client-urls ${GLOBAL_ETCD_SERVER} --listen-peer-urls ${GLOBAL_ETCD_PEER_SERVER} \
+  --initial-advertise-peer-urls ${GLOBAL_ETCD_PEER_SERVER} --initial-cluster global=${GLOBAL_ETCD_PEER_SERVER} \
+  --initial-cluster-token=${TOKEN} --initial-cluster-state=new \
+  ${OTHER_ETCD_FLAGS}
+```
+
+2. Associate vtctld to the etcd service and initialize global topology values
```
mattlord: Nit, but I would say tell vtctld to use the global topology server (there's not really anything to write yet as we have no keyspaces or cells). Whether or not the global root gets created head of time is implementation dependent (since it's really only a key prefix).

content/en/docs/reference/features/topology-service.md
```diff
@@ -418,7 +418,144 @@ use a long poll whose duration is set by the
...
+vtctld -topo_implementation=etcd2 -topo_global_server_address=${GLOBAL_ETCD_SERVER} \
+  -topo_global_root=/vitess/global -port=15000 -grpc_port=15999 -service_map='grpc-vtctl' \
+  ${OTHER_VTCTLD_FLAGS}
+```
+
+3. Create a local etcd service to store our cell information
+
+``` sh
+CELL_NAME="US_EAST"
+CELL_TOKEN="SOMETHING_UNIQ_HERE_TOO"
```
mattlord: This is just a unique identifier for participating processes, so I'd suggest cell name.

content/en/docs/reference/features/topology-service.md
```diff
@@ -418,7 +418,144 @@ use a long poll whose duration is set by the
...
+CELL_ETCD_PEER_SERVER="http://${CELL_ETCD_IP}:2380"
+
+etcd --enable-v2=true --data-dir ${VTDATAROOT}/etcd/${CELL_NAME} --listen-client-urls ${CELL_ETCD_SERVER} \
+  --name=${CELL_NAME} --advertise-client-urls ${CELL_ETCD_SERVER} --listen-peer-urls ${CELL_ETCD_PEER_SERVER} \
+  --initial-advertise-peer-urls ${CELL_ETCD_PEER_SERVER} --initial-cluster ${CELL_NAME}=${CELL_ETCD_PEER_SERVER} \
+  --initial-cluster-token=${CELL_TOKEN} --initial-cluster-state=new \
+  ${OTHER_ETCD_FLAGS}
+```
+
+4. Register the local etcd service to the global topology service using vtctl or
```
mattlord: Nit but I would say that we're defining the cell and its associated local topology service in the global topology service.

We are also slowly moving away from vtctl to vtctlclient and the only difference between them here is whether you specify the VTCTLD endpoint info or the TOPOSERVER endpoint info. So I'd just use the vtctlclient example to simplify things.

content/en/docs/reference/features/topology-service.md
```diff
@@ -418,7 +418,144 @@ use a long poll whose duration is set by the
...
+vtgate ${TOPOLOGY} -cell=${CELL_NAME} -cells_to_watch=${CELL_NAME} -port=15001 -grpc_port=15991 \
+-mysql_server_port=25306 -mysql_auth_server_impl=none -service_map='grpc-vtgateservice' \
+-tablet_types_to_wait PRIMARY,REPLICA \
+${OTHER_VTGATE_FLAGS}
+```
+
+```sh
+# vttablet implementation
+TOPOLOGY="-topo_implementation etcd2 -topo_global_server_address ${GLOBAL_ETCD_SERVER} -topo_global_root /vitess/global"
+TAB_NAME=${CELL_NAME}-100
```
mattlord: This is where you define the cell for the vttablet, FWIW. Whereas vtgate has an explicit cell flag.

mattlord: I would also call it `TABLET_ALIAS` as that's what it is in the product, and it consists of two parts:
${CELL}-${TABLET_UID} (UID must be unique within the cell and the alias must be unique in the cluster)

content/en/docs/reference/features/topology-service.md
```diff
@@ -418,7 +418,144 @@ use a long poll whose duration is set by the
...
+KEYSPACE=CustomerInfo
+
+vttablet ${TOPOLOGY} -tablet-path=${TAB_NAME} -tablet_dir=${VTDATAROOT}/${TAB_NAME} \
+  -mycnf-file=${VTDATAROOT}/${TAB_NAME}/my.cnf -init_keyspace=${KEYSPACE} -enable_semi_sync=true \
+  -init_shard=0 -init_tablet_type=replica -port=15100 -grpc_port=16100 \
+  -service_map='grpc-queryservice,grpc-tabletmanager,grpc-updatestream' \
+  ${OTHER_VTTABLET_FLAGS}
+```
+
+6. You can repeat steps 3 through 5 above to create each cell as needed. If you
```
mattlord, line 569: RebuildKeyspaceGraph is for vtgates and their serving graph (exposed via mysql> `SHOW VITESS_TABLETS`)

RebuildVSchemaGraph will propagate the vschema from the global topology service to the local topology services in their SrvVSchema. Not sure how best to express this here w/o adding confusing, but noting it. We could at least know that we're ensuring the local topology servers are up to date by propagating the latest info in the global service.

mattlord (changes requested):
Looking good! I left some minor comments and suggestions, let me know when you want me to do another pass.

Thank you! ❤️

content/en/docs/reference/features/topology-service.md
FancyFane: I agree, in my mind I was trying to think of what we call the backend service (etcd, zk, and consul) BEFORE it gets stamped/configured with the global topology information for Vitess. I'll work on cleaning up things a bit more here.

content/en/docs/reference/features/topology-service.md
FancyFane: Thank you Matt, when looking at this I could see both vtctl and the vtctlclient as similar in nature. Knowing the context of where Vitess was and where it is going I'll keep the vtctlclient examples.

content/en/docs/reference/features/topology-service.md
FancyFane: When I was working with discrete global/local instances, I came across the issue of not having up-to-date serving graph information on the vtgate, and noted Sugu's comment on this [blog entry](https://vitess.io/blog/2020-04-27-life-of-a-cluster/#understanding-keyspaces-and-shards). I figured it may be better to make a note of it here as well, in the event others had this problem.

content/en/docs/reference/features/topology-service.md
FancyFane: @mattlord can you let me know what you think of the new wording here? I'm trying to be explicitly clear that we need a keystore service (etcd, zk, or consul) BEFORE we can have topology service. Can you let me know your thoughts here?

I tried to follow the use of the word `keystore` with `(etcd, zookeeper, consul)`

```## Running multi cell environments

When running an environment with multiple cells, it is essential to first create
and configure your global topology service. Then define each local topology
service to the global topology. As mentioned perviously, the global
and local topology service can reside on the same or sperate keystore
(etcd, zookeeper, or consul) implementations. At a higher level overview:

* Create or locate an existing keystore (etcd, zookeeper, or consul) for the
  global keystore service.
* Use vtctl client commands to initialize the global topology service,
  providing the keystore implementation, address, and root directory.
  NOTE: for best practices the root dir should be set to `/vitess/global`.
* (Optional) For each cell create a keystore (etcd, zookeeper, or consul) for
  the local topology service. This step is optional as you may use the existing
  keystore implementation used by the global topology service. NOTE: the 
  keystore technologies must match. For example, if you are using zookeeper for
  your global topology service then you must use zookeeper for your local
  topology service. You can not mix and match.
* For each cell, using vtctl client commands, define the local topology service
  with the global topology service. This is done by providing the three
  elements: the cell name, the local keystore address/port, and the root
  directory. NOTE: for best practices the local root dir should be set to
  `/vitess/${CELL_NAME}` where ${CELL_NAME} is the location of the cell;
  `us-west-2`, `us-east-1`, etc.
* When starting vtgate/vttablet services, you will provide the global topology
  information, as well as the cell name. The global topology service will do
  a lookup on the cell name, to find and provide the local topology details back
  to the vtgate/vttablet process.

FancyFane: Resolved by simply referring to these services as "instance" when a general word was needed. This was done previously in the document.

content/en/docs/reference/features/topology-service.md
FancyFane: s/register/define/g

content/en/docs/reference/features/topology-service.md
FancyFane: Updated to say "vtctl client commands" and leave open for end user to determine which one to use. Example commands later in the document will just use `vtctlclient`.

content/en/docs/reference/features/topology-service.md
FancyFane: Done. Also did a "Ctrl+F" search for "keystore" to ensure that term is not being introduced in this document. Thank you for the clarification here.

content/en/docs/reference/features/topology-service.md
FancyFane: I agree, and have added example names based on AWS locations to give the reader some hints here.

content/en/docs/reference/features/topology-service.md
FancyFane: Upon reading your notes and reflecting a bit more, I've separated this out, so we can discuss vtgate and vttablet independently. Also, added a note about the potential for vtgate to watch more than one cell and thus more than one local topology service.

FancyFane: I did not include a note about the `MoveTables` operation as it may be outside the scope of this document. EDIT: But still good information to be aware of, thank you.

content/en/docs/reference/features/topology-service.md
FancyFane: I wrote it this way from local testing and seeing vtctld updating values once this command is ran. After you've provided clarification I re-wrote this step, to just say vtctld is simply using the global topology service created in step 1.

content/en/docs/reference/features/topology-service.md
FancyFane: Not a bad idea, as cell names should be unique. Implemented.

content/en/docs/reference/features/topology-service.md
FancyFane: s/register/define/g

Also, removed the `vtctl` example, so we can steer our readers towards using `vtctlclient`.

content/en/docs/reference/features/topology-service.md
FancyFane: Thank you, I've made this change as well as add some additional documentation on these variables. I've split out step 5 into two step 5s so we can clarify some of the differences between vtgate and vttablet.

FancyFane:
Upon editing this document, I've also realized we have several variables called `TOPOLOGY` while discussing both local and global topology services. To help ensure clarity I've renamed these variables to `GLOBAL_TOPOLOGY`.

mattlord (approved):
Thanks, @FancyFane ! ❤️

PR #859, opened 2021-10-31
Clean up stale/broken JavaScript and remove jQuery

doeg:
This fixes the errors getting thrown by the [tocify plugin](http://gregfranko.com/jquery.tocify.js/):

image

...and by "fixes", I mean "delete a bunch of code". :) 

On investigation, it turns out we can completely remove jQuery and save ~115 KB of production assets. 🎉 Anything that's currently "using" it is broken or unnecessary. I've added some inline explanations.

PR #882, opened 2021-11-17
Rewrite expand.html partial to use details element

doeg:
This rewrites the expand.html partial to use the browser native `<details>` element. (Easily my favourite HTML element of all time. :sunglasses: Love it so much.)

Here's a before/after to show what it looks like:

Screen Shot 2021-11-17 at 10 00 34 AM

Screen Shot 2021-11-17 at 9 59 47 AM

This fixes the regression found by @mattlord (thank you!) from #859, which removed jQuery and broke the inline `onclick` handler in the previous implementation.

One small browser compatibility note: `<details>` is not supported by Internet Explorer, however it fails "open" (the content is displayed but there's no toggle functionality). So, I didn't bother adding a polyfill. Here's what it looks like in IE:

image

### DIFF
--- a/assets/sass/style.sass
+++ b/assets/sass/style.sass
@@ -237,3 +237,6 @@ h1, h2, h3, h4, h5, h6
     height: $navbar-anchor-offset
     visibility: hidden
     pointer-events: none
+
+details summary
+  cursor: pointer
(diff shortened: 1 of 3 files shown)


mattlord (approved):
Thank you!

deepthi (approved):

PR #881, opened 2021-11-17
Update default collations docs

frouioui:
## Description

This pull request updates the documentation related to default collation after the merge of https://github.com/vitessio/vitess/pull/9248.

PR #883, opened 2021-11-17
docs: add collations.md

vmg:
Let's write some proper documentation for this feature, since collations are not easy to understand.

cc @frouioui @systay @deepthi


frouioui:
We can close #881

frouioui (approved):
This is very good @vmg! Liked it a lot

shlomi-noach:
Late for the party, but agreed -- awesome writeup @vmg !

PR #934, opened 2022-01-06
fixing links

hallaroo:
Going through the links broken by the versioning and getting them fix. This is likely just round 1 of a few.


deepthi (approved):
These look correct to me. Using relative paths means that docs in one release only refer to docs in the same release 👍

PR #938, opened 2022-01-10
Fixes for v13 links 

hallaroo:
Fixes for the broken links in v13 after the change to versioned docs. Still need to do v12 and v11.


deepthi (approved):
I'm assuming all the links have been checked 🙂

PR #948, opened 2022-01-22
Document the usage of PlannedReparentShard with init_db.sql file

GuptaManan100:
Release 13 onwards, we will be able to use PlannedReparentShard to initialise the cluster even with a `init_db.sql` file as long as no errant GTIDs are produced. This PR documents that scenario.

### DIFF
--- a/content/en/docs/13.0/user-guides/configuration-basic/initialize-shard-primary.md
+++ b/content/en/docs/13.0/user-guides/configuration-basic/initialize-shard-primary.md
@@ -7,7 +7,9 @@ The [PlannedReparentShard](../../configuration-advanced/reparenting/#plannedrepa
 
 The InitShardPrimary step can also be used to do the same operation. However, it is a destructive command and should only be used by advanced users. This command copies over the `executed_gtid_set` from the primary to the replica which can break replication if the user isn't careful. 
 
-InitShardPrimary must be used in case the users are using a custom `init_db.sql` file in their setup which creates errant GTIDs on the vttablets. If the users have `SET sql_log_bin = 0` in the beginning of their file (like in the example `init_db.sql` file), then PlannedReparentShard will work as expected and it should be used.
+{{< info >}}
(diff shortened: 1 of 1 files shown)


content/en/docs/13.0/user-guides/configuration-basic/initialize-shard-primary.md
```diff
@@ -7,6 +7,8 @@ The [PlannedReparentShard](../../configuration-advanced/reparenting/#plannedrepa
 
 The InitShardPrimary step can also be used to do the same operation. However, it is a destructive command and should only be used by advanced users. This command copies over the `executed_gtid_set` from the primary to the replica which can break replication if the user isn't careful. 
 
+InitShardPrimary must be used in case the users are using a custom `init_db.sql` file in their setup which creates errant GTIDs on the vttablets. If the users have `SET sql_log_bin = 0` in the beginning of their file (like in the example `init_db.sql` file), then PlannedReparentShard will work as expected and it should be used.
```
hallaroo: InitShardPrimary must be used when you add a custom `init_db.sql` file that creates errant GTIDs on vttablets to your setup. If you have `SET sql_log_bin = 0` at the beginning of your file (like in the example `init_db.sql` file), then PlannedReparentShard will work as expected and it should be used.

hallaroo (changes requested):
Adjusting a bit for readability.

content/en/docs/13.0/user-guides/configuration-basic/initialize-shard-primary.md
deepthi: Let's make this absolutely clear.. Something like
```
If using a custom `init_db.sql` that omits `SET sql_log_bin = 0`, then InitShardPrimary must be used instead of PlannedReparentShard.
```

hallaroo (approved):
LGTM!

PR #957, opened 2022-01-30
Document -max_replication_lag_allowed flag

rohit-nayak-ps:
Document new flag implemented in https://github.com/vitessio/vitess/pull/9538

Relevant change here: https://deploy-preview-957--vitess.netlify.app/docs/13.0/reference/vreplication/switchtraffic/#-max_transaction_lag_allowed

### DIFF
--- a/content/en/docs/13.0/reference/vreplication/switchtraffic.md
+++ b/content/en/docs/13.0/reference/vreplication/switchtraffic.md
@@ -69,6 +69,21 @@ If set to false these reverse replication streams will not be created and you wi
 
 </div>
 
+#### -max_transaction_lag_allowed
+**optional**\
+**default**  the value used for `-timeout`
+
+<div class="cmd">
+
+While switching traffic ensure that the VReplication lag for the workflow is less than this duration, otherwise
+report an error and don't switch. VReplication lag is the estimated maximum lag across workflow streams between the last event seen at the source and the event processed by the target. Usually, when VReplication has caught up, this should be very small (under a second).
+
+While switching write traffic, we temporarily make the source databases read-only, and wait for the targets to catchup. The application will be down for this time: writes will error out. While switching write traffic, this flag can ensure that you only switch traffic if the current lag is small limiting downtime.
+
+This can also be useful if only reads are being switched: if the lag is too high the app might be reading stale data.
+
+</div>
+
 #### -dry-run
 **optional**\
 **default** false
--- a/content/en/docs/13.0/reference/vreplication/switchtraffic.md
+++ b/content/en/docs/13.0/reference/vreplication/switchtraffic.md
@@ -75,12 +75,11 @@ If set to false these reverse replication streams will not be created and you wi
 
 <div class="cmd">
 
-While switching traffic ensure that the VReplication lag for the workflow is less than this duration, otherwise
-report an error and don't switch. VReplication lag is the estimated maximum lag across workflow streams between the last event seen at the source and the event processed by the target (which would be a heartbeat event if we're fully caught up). Usually, when VReplication has caught up, this should be very small (under a second).
+While switching traffic ensure that the VReplication lag for the workflow is less than this duration, otherwise report an error and don't attempt the switch. The calculated VReplication lag is the estimated maximum lag across workflow streams between the last event seen at the source and the last event processed by the target (which would be a heartbeat event if we're fully caught up). Usually, when VReplication has caught up, this lag should be very small (under a second).
 
-While switching write traffic, we temporarily make the source databases read-only, and wait for the targets to catchup. The application will be down for this time: writes will error out. While switching write traffic, this flag can ensure that you only switch traffic if the current lag is small limiting write-unavailability.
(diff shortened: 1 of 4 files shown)


mattlord:
As noted in the PR, I think we should use `-max_[v]replication_lag_allowed` as the name.

content/en/docs/13.0/reference/vreplication/switchtraffic.md
```diff
@@ -69,6 +69,21 @@ If set to false these reverse replication streams will not be created and you wi
 
 </div>
 
+#### -max_transaction_lag_allowed
+**optional**\
+**default**  the value used for `-timeout`
+
+<div class="cmd">
+
+While switching traffic ensure that the VReplication lag for the workflow is less than this duration, otherwise
```
mattlord: it's number of seconds, right? IMO, saying duration here may be confusing as many flags take a golang duration value like "5s" but here it would just be 5.

content/en/docs/13.0/reference/vreplication/switchtraffic.md
```diff
@@ -69,6 +69,21 @@ If set to false these reverse replication streams will not be created and you wi
...
+report an error and don't switch. VReplication lag is the estimated maximum lag across workflow streams between the last event seen at the source and the event processed by the target. Usually, when VReplication has caught up, this should be very small (under a second).
```
mattlord: IMO we should mention the heartbeat as well, e.g.: 
```
VReplication lag is the estimated maximum lag across workflow streams between the last event seen at the source and the last event processed by the target (which would be a heartbeat event if we're fully caught up).
```

content/en/docs/13.0/reference/vreplication/switchtraffic.md
```diff
@@ -69,6 +69,21 @@ If set to false these reverse replication streams will not be created and you wi
...
+
+While switching write traffic, we temporarily make the source databases read-only, and wait for the targets to catchup. The application will be down for this time: writes will error out. While switching write traffic, this flag can ensure that you only switch traffic if the current lag is small limiting downtime.
```
mattlord: I would say ~ "write unavailability" instead of downtime.

content/en/docs/13.0/reference/vreplication/switchtraffic.md
```diff
@@ -69,6 +69,21 @@ If set to false these reverse replication streams will not be created and you wi
...
+
+This can also be useful if only reads are being switched: if the lag is too high the app might be reading stale data.
```
mattlord: I would say something like:
```
While switching read traffic this can be used to set an approximate upper bound on how stale reads will be against the replica tablets when using `@replica` shard targeting.
```

mattlord (changes requested):

content/en/docs/13.0/reference/vreplication/switchtraffic.md
rohit-nayak-ps: It is a duration `maxReplicationLagAllowed := subFlags.Duration("max_replication_lag_allowed"`

content/en/docs/13.0/reference/vreplication/switchtraffic.md
rohit-nayak-ps: fixed as suggested

content/en/docs/13.0/reference/vreplication/switchtraffic.md
rohit-nayak-ps: fixed as suggested

mattlord (approved):

content/en/docs/13.0/reference/vreplication/switchtraffic.md
mattlord: Ah, OK. So instead it's this that I think is confusing/misleading: https://github.com/vitessio/vitess/blob/6443bb502885dda91af6b5337e7bc682cb2861d9/go/vt/vtctl/vtctl.go#L2299

the "in seconds" part. No?

content/en/docs/13.0/reference/vreplication/switchtraffic.md
```diff
@@ -69,6 +69,21 @@ If set to false these reverse replication streams will not be created and you wi
 
 </div>
 
+#### -max_transaction_lag_allowed
```
mattlord: We need to correct this to: `max_replication_lag_allowed`

mattlord (changes requested):
In addition to updating the flag name to `max_replication_lag_allowed`, could you please copy these changes to the new 14.0 versions of the pages?
- content/en/docs/14.0/reference/vreplication/flags.md
- content/en/docs/14.0/reference/vreplication/switchtraffic.md 

Those pages started as carbon copies of the 13.0 ones, so we can just copy them as-is to the 14.0 directory. 

Thank you!

content/en/docs/13.0/reference/vreplication/switchtraffic.md
rohit-nayak-ps: Yes, the comment for the flag is incorrect, will fix in my next PR.

mattlord (approved):
Thanks!

PR #962, opened 2022-02-01
Online DDL doc updates: -migration_context and -json

shlomi-noach:
### This should make it to v13 release

This documentation update follows up on:

- https://github.com/vitessio/vitess/pull/9568
- https://github.com/vitessio/vitess/pull/9572

we prefer to use `-migration_context` from now on, so the docs use this flag now.


hallaroo (approved):
LGTM. But with one additional thing. I think we'd need to update the v14 docs as well with these changes.

shlomi-noach:
> I think we'd need to update the v14 docs as well with these changes.

Oh, right! OK, I'll add the changes to this PR

shlomi-noach:
Applied same changes in `14.0`!

That'll teach me to submit a PR just before a new version rolls out.

PR #987, opened 2022-02-23
Correct mysql command for connecting to vtgate

madhur:
The mysql command stated in docs is incorrect , it should be

```
 mysql -h 127.0.0.1 -P 15306 --user=mysql_user --password=mysql_password
```

### DIFF
--- a/content/en/docs/11.0/user-guides/configuration-basic/vtgate.md
+++ b/content/en/docs/11.0/user-guides/configuration-basic/vtgate.md
@@ -87,7 +87,7 @@ If vtgates cannot connect to one of the vttablets it discovered from the topo, o
 You can verify that the vtgates came up successfully by using the MySQL client:
 
 ```text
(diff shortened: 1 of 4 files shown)


mattlord:
Good catch! This change is fine, alternatively we could have a space: `-u user -p passwd`. Either is fine.

Once you [fix the DCO as described here](https://github.com/vitessio/website/pull/987/checks?check_run_id=5308318099) then I'll approve and merge. 

THANK YOU!

madhur:
> Good catch! This change is fine, alternatively we could have a space: `-u user -p passwd`. Either is fine.
> 
> Once you [fix the DCO as described here](https://github.com/vitessio/website/pull/987/checks?check_run_id=5308318099) then I'll approve and merge.
> 
> THANK YOU!

Thanks @mattlord . Fixed the DCO

mattlord (approved):
Thanks again!

PR #998, opened 2022-03-09
Add tablet_types_to_wait parameter

madhur:
Found this to be mandatory parameter while running vtgate on v12


deepthi (commented):
Thank you for the contribution and apologies for the late review.
This looks fine. Do you mind adding the same to 13.0 and 14.0 docs?

madhur:
> Thank you for the contribution and apologies for the late review. This looks fine. Do you mind adding the same to 13.0 and 14.0 docs?

Done

deepthi (approved):
LGTM. Thank you!

PR #1007, opened 2022-04-07
Add initial Architecture docs for VTAdmin

doeg:
Signed-off-by: Sara Bee <855595+doeg@users.noreply.github.com>

Branch preview: https://deploy-preview-1007--vitess.netlify.app/docs/14.0/reference/vtadmin/architecture/

A couple of notes:

- When we settle on a mergeable version of these docs, I'll "backport" (copy and paste, lol) the docs from 14.0 to the 13.0 release section. (Won't bother for 12.0 unless folks have strong feelings.) 

- I think that documentation on service discovery would fit nicely in this "Architecture" doc. I didn't do that here, since @notfelineit mentioned an interest in doing it. :) 

- The mermaid.js charts are... not very cute. I'll probably style 'em up in a separate PR, if we decide we want to use mermaid.js. Definitely open to opinions on whether we want to keep it! I like it a lot so far, since it's real nice to define graphs inline like that. (Consistent! Readable! Source controlled!) It does get a little messy when it comes to complicated charts, but tbh it's a reasonable forcing function for making non-complicated + more readable charts. :') (I also played around with the built-in GoAT ASCII diagrams but they're not super accessible IMO.) 

Screenshot of not-yet-cute mermaid.js diagrams for posterity:

screencapture-localhost-1313-docs-14-0-reference-vtadmin-architecture-2022-04-08-10_04_44


doeg:
I don't have permissions to add folks as reviewers, so I'll cc y'all here: @ajm188 @deepthi @notfelineit. 🐈‍⬛  Would love any + all feedback you have!

layouts/partials/javascript.html
```diff
@@ -14,3 +14,13 @@
   debug: false
 });
 </script>
+
+<script src="https://cdn.jsdelivr.net/npm/mermaid/dist/mermaid.min.js"></script>
```
doeg, line 18: I'm being kinda lazy by using a CDN link for this! 😎 But also it's not the worst!!! (I get SO easily derailed from writing docs so I didn't do that here -- I'll fix it when I come back around to make the mermaid diagrams cute, if we decide to keep it.)

content/en/docs/14.0/reference/vtadmin/architecture.md
```diff
@@ -0,0 +1,91 @@
...
+        vtgate_ap <--> topology_ap
+        vtctld_ap <--> topology_ap
+    end
+```
+
+# The Life Cycle of a VTAdmin Request
+
+To understand how data moves through VTAdmin, let's look at life cycle of a typical request: fetching a list of all of the schemas. We'll use the single-cluster ["Get Started"](../../get-started/) environment as an example:
+
+<img src="../img/schemas.png" width="960" alt="A list of schemas in VTAdmin" />
```
doeg, line 62: I want to investigate some GitHub Actions hijinks to keep these screenshots in sync with the actual VTAdmin codebase. (e.g., generating screenshots on vtadmin-web updates + opening a PR with updated images in this repo... that's the dream, anyway.)

layouts/partials/javascript.html
doeg: Ok so after investigating [how Hugo does JavaScript builds](https://gohugo.io/hugo-pipes/js/), I lost about 80% of my HP and think we should just use the CDN until someone really _really_ wants to get in the weeds of our JavaScript pipeline. (A CDN build is probably faster anyway.)

ajm188: I think you want to put it in `/assets` (https://gohugo.io/hugo-pipes/bundling/), unless there's a licensing issue around doing that.

ajm188 (approved):
lgtm!

layouts/partials/javascript.html
doeg: Yeah, so what I wanted to do was to install the package with npm and then import it in the usual way (`import mermaid from 'mermaidjs'` or whatever). I think Hugo also supports [symlinking/"mounting"](https://discourse.gohugo.io/t/importing-js-from-node-packages/23545/3) `node_modules`, so... it's possible, just not straightforward, and I don't want to reduce my documentation """"momentum""".

doeg:
@ajm188 or @deepthi, would you mind merging this for me? 🙇

doeg:
Oh WAIT I have to copy everything into v13... I knew I'd forget that, lol. Please hold. ☎️

doeg:
Ok it is done 😌 and ready to merge whenever.

PR #1015, opened 2022-04-24
Update message table schema for efficient poller and w/o potential message loss

mattlord:
This updates the docs based on the work done in: https://github.com/vitessio/vitess/pull/10132

The potential for loss of messages — or more accurately messages never sent — comes from the `time_next bigint` column definition that we had in the docs, which is effectively `time_next bigint DEFAULT NULL`.  If the `time_next` value is NULL, then [it *never* gets read from the underlying message table in the poller / results streamer query](https://github.com/vitessio/vitess/blob/c9853797bda8752202583695133c41a25d390930/go/vt/vttablet/tabletserver/messager/message_manager.go#L254-L256) because `<` is *not* NULL safe and that `< <time_next_val>` clause will never match those rows where the value is NULL (unknown). So while rows INSERTed with `time_next` being NULL _while the binlog streamer is running_ will catch these rows and process them, any rows INSERTed _while the binlog streamer is not running, or the cache is full, will be lost._ 

We were already [using a `DEFAULT 0` clause for the `time_next` column in our tests](https://github.com/vitessio/vitess/pull/10132/commits/1d3cbdcbec9a7db8aed3ac94eb6210131a68ffe1), but this was not done in the docs.

### DIFF
--- a/content/en/docs/11.0/reference/features/messaging.md
+++ b/content/en/docs/11.0/reference/features/messaging.md
@@ -51,18 +51,24 @@ message table must be created like this:
 
 ```sql
 create table my_message(
-  time_scheduled bigint,
-  id bigint,
-  time_next bigint DEFAULT 0,
-  epoch bigint,
-  time_created bigint,
-  time_acked bigint,
-  message varchar(128),
-  priority tinyint NOT NULL DEFAULT 0,
-  primary key(time_scheduled, id),
-  unique index id_idx(id),
+  # required columns
+  id bigint NOT NULL COMMENT 'often an event id, can also be auto-increment or a sequence',
+  priority tinyint NOT NULL DEFAULT '50' COMMENT 'lower number priorities process first',
+  epoch bigint NOT NULL DEFAULT '0' COMMENT 'Vitess increments this each time it sends a message, and is used for incremental backoff doubling',
+  time_next bigint DEFAULT 0 COMMENT 'the earliest time the message will be sent in epoch nanoseconds. Must be null if time_acked is set',
+  time_acked bigint DEFAULT NULL COMMENT 'the time the message was acked in epoch nanoseconds. Must be null if time_next is set',
+
+  # add as many custom fields here as required
+  # optional - these are suggestions
+  tenant_id bigint,
+  message json,
+
+  # required indexes
+  primary key(id),
   index poller_idx(time_acked, priority, time_next desc)
-) comment 'vitess_message,vt_ack_wait=30,vt_purge_after=86400,vt_batch_size=10,vt_cache_size=10000,vt_poller_interval=30'
+
+  # add any secondary indexes or foreign keys - no restrictions
+) comment 'vitess_message,vt_min_backoff=30,vt_max_backoff=3600,vt_ack_wait=30,vt_purge_after=86400,vt_batch_size=10,vt_cache_size=10000,vt_poller_interval=30'
 ```
 
 The application-related columns are as follows:
--- a/content/en/docs/11.0/reference/features/messaging.md
+++ b/content/en/docs/11.0/reference/features/messaging.md
@@ -166,8 +166,8 @@ exceeds the time period specified by `vt_purge_after`.
 ## Advanced usage
 
 The `MessageAck` functionality is currently an gRPC API call and cannot be used
-inside a transaction or more generally from the SQL interface. However, you can
-manually ack messages using a regular DML query like this:
+from the SQL interface. However, you can manually ack messages using a regular
+DML query like this:
 
 ```sql
 update my_message set time_acked = :time_acked, time_next = null where id in ::ids and time_acked is null
--- a/content/en/docs/11.0/reference/features/messaging.md
+++ b/content/en/docs/11.0/reference/features/messaging.md
@@ -53,15 +53,15 @@ message table must be created like this:
 create table my_message(
   time_scheduled bigint,
   id bigint,
-  time_next bigint,
+  time_next bigint DEFAULT 0,
   epoch bigint,
   time_created bigint,
   time_acked bigint,
   message varchar(128),
-  priority tinyint NOT NULL DEFAULT '0',
+  priority tinyint NOT NULL DEFAULT 0,
   primary key(time_scheduled, id),
   unique index id_idx(id),
-  index next_idx(priority asc, time_next desc)
+  index poller_idx(time_acked, priority, time_next desc)
 ) comment 'vitess_message,vt_ack_wait=30,vt_purge_after=86400,vt_batch_size=10,vt_cache_size=10000,vt_poller_interval=30'
 ```
 
--- a/content/en/docs/11.0/reference/features/messaging.md
+++ b/content/en/docs/11.0/reference/features/messaging.md
@@ -44,7 +44,8 @@ Messages are not a good fit for the following use cases:
 ## Creating a message table
 
 The current implementation requires a base fixed schema with properties defined
-using specific table COMMENT directives. The message table format is as follows:
+using Vitess specific table `COMMENT` directives. The message table format is as
+follows:
 
 ```sql
 create table my_message(
@@ -81,7 +82,7 @@ The comment section specifies additional configuration parameters. The fields
 are as follows:
 
 * `vitess_message`: Indicates that this is a message table.
-* `vt_min_backoff=30`, `vt_max_backoff=3600`: set bounds, in seconds, on exponential
+* `vt_min_backoff=30`, `vt_max_backoff=3600`: Set bounds, in seconds, on exponential
   backoff for message retries.
 * `vt_ack_wait=30`: Wait for 30 seconds for the *first* message send to be acked.
   If one is not received within this time frame, the message will be resent.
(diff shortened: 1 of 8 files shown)


derekperkins:
`time_scheduled` and `time_created` are no longer populated by Vitess, which impacts the default table. Here's an updated suggestion with comments.

```sql
create table my_message(
  # required columns
  id bigint NOT NULL COMMENT 'often an event id, can also be auto-increment or a sequence',
  priority tinyint NOT NULL DEFAULT '50' COMMENT 'lower number priorities process first',
  epoch bigint NOT NULL DEFAULT '0' COMMENT 'Vitess increments this each time it sends a message, and is used for incremental backoff doubling',
  time_next bigint DEFAULT 0 COMMENT 'the earliest time the message will be sent in epoch nanoseconds. Must be null if time_acked is set',
  time_acked bigint DEFAULT NULL COMMENT 'the time the message was acked in epoch nanoseconds. Must be null if time_next is set',

  # add as many custom fields here as required
  # optional - these are suggestions
  tenant_id bigint,
  message json,

  # required indexes
  primary key(id),
  index poller_idx(time_acked, priority, time_next desc)

  # add any secondary indexes or foreign keys - no restrictions
) comment 'vitess_message,vt_min_backoff=30,vt_max_backoff=3600,vt_ack_wait=30,vt_purge_after=86400,vt_batch_size=10,vt_cache_size=10000,vt_poller_interval=30'
```

derekperkins:
We don't have to address it in the PR, but I wonder if we could/should introduce a check constraint for anyone >= 8.0.16 that enforces the mutual exclusivity of `time_next` and `time_acked`

mattlord:
@derekperkins could you please review the new page whenever you have time? I went through the entire thing and updated it. All of them are the same, so you can look at this one: https://deploy-preview-1015--vitess.netlify.app/docs/14.0/reference/features/messaging/

Thanks!

content/en/docs/11.0/reference/features/messaging.md
```diff
@@ -34,102 +33,102 @@ Messages are good for:
...
+  # required columns
+  id bigint NOT NULL COMMENT 'often an event id, can also be auto-increment or a sequence',
+  priority tinyint NOT NULL DEFAULT '50' COMMENT 'lower number priorities process first',
+  epoch bigint NOT NULL DEFAULT '0' COMMENT 'Vitess increments this each time it sends the message, and is used for incremental backoff doubling',
+  time_next bigint DEFAULT 0 COMMENT 'the earliest time the message will be sent in epoch nanoseconds. Must be null if time_acked is set',
+  time_acked bigint DEFAULT NULL COMMENT 'the time the message was acked in epoch nanoseconds. Must be null if time_next is set',
+
+  # add as many custom fields here as required
+  # optional - these are suggestions
+  tenant_id bigint COMMENT 'offers a nice way to segment your messages',
```
derekperkins, line 61: Should we say "shard" instead of "segment"?

content/en/docs/11.0/reference/features/messaging.md
```diff
@@ -34,102 +33,102 @@ Messages are good for:
...
 transaction as it consumes memory on the VTTablet side. At the time of commit,
 memory permitting, all messages are instantly enqueued to be sent.
 
-Messages can also be created to be sent in the future:
-
-```sql
-insert into my_message(id, message, time_scheduled) values(1, 'hello world', :future_time)
-```
-
-`future_time` must be the unix time expressed in nanoseconds.
```
derekperkins: This is still possible - replace `time_scheduled` with `time_next`

content/en/docs/11.0/reference/features/messaging.md
```diff
@@ -153,9 +153,10 @@ Once a message is successfully acked, it will never be resent.
 
 ## Exponential backoff
 
-A message that was successfully sent will wait for the specified ack wait time.
-If no ack is received by then, it will be resent. The next attempt will be 2x
-the previous wait, and this delay is doubled for every attempt.
+For a message that was successfully sent we will wait for the specified `vt_ack_wait`
+time. If no ack is received by then, it will be resent. The next attempt will be 2x
+the previous wait, and this delay is doubled for every attempt (with some added
+jitter to avoid thundering herds).
```
derekperkins: Should probably add details for `vt_min_backoff` and `vt_max_backoff` here.

Also, maybe we should specify that "some added jitter" is "up to 33% jitter"

content/en/docs/11.0/reference/features/messaging.md
```diff
@@ -164,50 +165,38 @@ exceeds the time period specified by `vt_purge_after`.
...
+update my_message set priority = :priority, time_next = :time_next, epoch = :epoch where id in ::ids and time_acked is null
 ```
 
 This comes in handy if a bunch of messages had chronic failures and got
 postponed to the distant future. If the root cause of the problem was fixed,
-the application could reschedule them to be delivered immediately. You can also
-optionally change the epoch. Lower epoch values increase the priority of the
-message and the back-off is less aggressive.
+the application could reschedule them to be delivered as soon as possible. You can
+also optionally change the priroity and or epoch. Lower priority and epoch values
```
derekperkins: s/priroity/priority

content/en/docs/11.0/reference/features/messaging.md
```diff
@@ -164,50 +165,38 @@ exceeds the time period specified by `vt_purge_after`.
...
+Here is a short list of possible limitations/improvements:
 
 * Proactive scheduling: Upcoming messages can be proactively scheduled for
   timely delivery instead of waiting for the next polling cycle.
 * Changed properties: Although the engine detects new message tables, it does
-  not refresh properties of an existing table.
-* A `SELECT` style syntax for subscribing to messages.
-* No rate limiting.
-* Usage of partitions for efficient purging.
+  not refresh the properties (such as `vt_ack_wait`) of an existing table.
```
derekperkins, line 210: I don't think this is true anymore, but I'm not sure

content/en/docs/11.0/reference/features/messaging.md
```diff
@@ -164,50 +165,38 @@ exceeds the time period specified by `vt_purge_after`.
...
-The message feature is currently in alpha, and can be improved. Here is the
-list of possible limitations/improvements:
+Here is a short list of possible limitations/improvements:
 
 * Proactive scheduling: Upcoming messages can be proactively scheduled for
   timely delivery instead of waiting for the next polling cycle.
 * Changed properties: Although the engine detects new message tables, it does
-  not refresh properties of an existing table.
-* A `SELECT` style syntax for subscribing to messages.
-* No rate limiting.
```
derekperkins, line 211: There still isn't any rate limiting

content/en/docs/11.0/reference/features/messaging.md
```diff
@@ -34,102 +33,102 @@ Messages are good for:
...
 variation can be allowed to achieve different performance trade-offs.
 
 The comment section specifies additional configuration parameters. The fields
 are as follows:
 
 * `vitess_message`: Indicates that this is a message table.
-* `vt_ack_wait=30`: Wait for 30s for the first message ack. If one is not
-  received, resend.
+* `vt_min_backoff=30`, `vt_max_backoff=3600`: Set bounds, in seconds, on exponential
+  backoff for message retries.
```
derekperkins: * `vt_min_backoff=30`: Set lower bound, in seconds, on exponential backoff for message retries. If not set, defaults to `vt_ack_wait` _(optional)_
* `vt_max_backoff=3600`: Set upper bound, in seconds, on exponential backoff for message retries. The default value is infinite backoff _(optional)_

content/en/docs/11.0/reference/features/messaging.md
derekperkins: Not sure if we want to get this technical, but here's a first stab at giving actual details

### Basic backoff equation:
```
jitteredBackoff = `vt_ack_wait` ^ `epoch` * (random value between 0.66 and 1.33)
actualBackoff = `vt_min_backoff` <= `jitteredBackoff` <= `vt_max_backoff`
```

### backoff query if `vt_max_backoff` is NOT set
```sql
update foo set time_next = 
  :time_now + :wait_time + IF(FLOOR((:min_backoff<<ifnull(epoch, 0)) * :jitter) < :min_backoff, :min_backoff, FLOOR((:min_backoff<<ifnull(epoch, 0)) * :jitter)), epoch = ifnull(epoch, 0)+1 where id in ::ids and time_acked is null
```

### backoff query if `vt_max_backoff` IS set
```sql
update foo set time_next = 
  :time_now + :wait_time + IF(FLOOR((:min_backoff<<ifnull(epoch, 0)) * :jitter) < :min_backoff, :min_backoff, IF(FLOOR((:min_backoff<<ifnull(epoch, 0)) * :jitter) > :max_backoff, :max_backoff, FLOOR((:min_backoff<<ifnull(epoch, 0)) * :jitter))), epoch = ifnull(epoch, 0)+1 where id in ::ids and time_acked is null
```

derekperkins: FYI, this is definitely one place where having to handle null epoch adds complexity. If we're ok enforcing non-null epoch, we could change this

derekperkins:
Adding `WHERE time_acked IS NULL` to the poller query actually solves some issues we've encountered in the past. If someone was requeuing messages after a fix, sometimes they would accidentally set `time_next` while the service set `time_acked`, in which case there ended up being an infinite loop of message retries. We should still probably note somewhere that the fields are designed to be mutually exclusive, but at least it's less of a footgun now.

content/en/docs/11.0/reference/features/messaging.md
mattlord: I feel like it could still be useful when not sharded. Some receivers could only process and ack messages for a given tenant. No?

content/en/docs/11.0/reference/features/messaging.md
mattlord: I would rather leave this as it is for now (we're already making a lot of changes) and not share this level of detail in the docs. For those interested, the source code is there.

content/en/docs/11.0/reference/features/messaging.md
derekperkins: We have lots of unsharded queues - tenant id just feels like the most common sharding id, though I understand your point. My first reaction on reading it was to wonder if "segment" was a specific concept for messaging. I don't feel super strongly about it. 

Once we're done with this, I want to blog about and/or add an opinionated design to the docs, with the specific fields and queries we use, so people don't have to architect from scratch.

derekperkins (approved):

content/en/docs/11.0/reference/features/messaging.md
mattlord: > We have lots of unsharded queues - tenant id just feels like the most common sharding id, though I understand your point. My first reaction on reading it was to wonder if "segment" was a specific concept for messaging. I don't feel super strongly about it.

I don't have a strong opinion here either. We can always revisit.

> Once we're done with this, I want to blog about and/or add an opinionated design to the docs, with the specific fields and queries we use, so people don't have to architect from scratch.

That would be awesome! ❤️

PR #1012, opened 2022-04-13
Add VTAdmin Operator's Guide

doeg:
Signed-off-by: Sara Bee <855595+doeg@users.noreply.github.com>

Deploy previews:
- [Operator's Guide](https://deploy-preview-1012--vitess.netlify.app/docs/14.0/reference/vtadmin/operators_guide/)
- [vtadmin-web reference](https://deploy-preview-1012--vitess.netlify.app/docs/14.0/reference/programs/vtadmin-web/)

This does not cover the changes proposed in https://github.com/vitessio/vitess/pull/10543, nor [single-component VTAdmin](https://github.com/vitessio/vitess/pull/10214) (which would simplify this a lot). Neither does it go into much detail about deployment specifics (Docker, k8s, etc.).

PR #1019, opened 2022-05-06
Document vtadmin flags in programs/reference/vtadmin

doeg:
Signed-off-by: Sara Bee <855595+doeg@users.noreply.github.com>

**🔬  Staging: https://deploy-preview-1019--vitess.netlify.app/docs/14.0/reference/programs/vtadmin/**

I extracted this from #1012, since these flag references tend to get out of date quickly so we may as well put a stake in the ground! (The [schema caching PR](https://github.com/vitessio/vitess/pull/10120) adds a few, as will [single-component vtadmin](https://github.com/vitessio/vitess/pull/10214).)

Two small things I changed after pilfering this from #1012:

- I had initially referred to the binary as `vtadmin-api` but it's actually called `vtadmin` (which makes even more sense in light of single-component vtadmin, so 🎉 for once naming worked out nicely!)

- I addressed [this comment](https://github.com/vitessio/website/pull/1012#discussion_r850480839) about wording (thank you!) 

**I intentionally did not add these to the 13.0 docs, since I'll have to go back and actually `git checkout` that tag + see if there are any inconsistencies.**

For full disclosure, I know we have some kind of tooling in vitessio/vitess to generate documentation, but in the interests of getting this out the door, I just wrote these cursed markdown tables by hand. 🤷

### DIFF
--- a/content/en/docs/14.0/reference/programs/vtadmin.md
+++ b/content/en/docs/14.0/reference/programs/vtadmin.md
@@ -21,7 +21,7 @@ One of `cluster`, `cluster-defaults`, or `cluster-config` file is required. Mult
 | Name | Definition |
 | -------- | --------- | 
 | `cluster` | Per-cluster configuration. Any values here take precedence over those defined by the `cluster-defaults` and/or `cluster-config` flags. | 
-| `cluster-config` | Path to a yaml cluster configuration; for reference, see the example [clusters.yaml](https://github.com/vitessio/vitess/blob/main/doc/vtadmin/clusters.yaml. | 
+| `cluster-config` | Path to a yaml cluster configuration; for reference, see the example [clusters.yaml](https://github.com/vitessio/vitess/blob/main/doc/vtadmin/clusters.yaml). | 
 | `cluster-defaults` | Default options for all clusters. |
 | `enable-dynamic-clusters` | Defaults to `false`. Whether to enable dynamic clusters that are set by request header cookies or gRPC metadata. | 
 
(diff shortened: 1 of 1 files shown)


content/en/docs/14.0/reference/programs/vtadmin.md
```diff
@@ -0,0 +1,77 @@
...
+
+These flags are provided to the `vtadmin` process. They are also referenced in [cmd/vtadmin/main.go](https://github.com/vitessio/vitess/blob/main/go/cmd/vtadmin/main.go).
+
+### Common flags
+
+| Name | Required | Type | Default | Definition |
+| -------- | --------- | --------- | --------- |--------- |
+| `addr` | **Required** | string | ":15000" | The address for `vtadmin` to serve on. |
+| `lame-duck-duration` | Optional | seconds | 5s | The length of the lame duck period at shutdown. |
+| `lmux-read-timeout` | Optional | seconds | 1s | How long to spend connection muxing. |
```
ajm188: technically, these are of type `time.Duration`, not "seconds" (i'm not _just_ being a pedant here (but SORRY), but this means you can pass a value of, say, `1h` as well)

ajm188 (approved):
one nitpick on the duration types, then feel free to merge!

content/en/docs/14.0/reference/programs/vtadmin.md
doeg: Oh thank you! Not pedantic at all!! I was unsure about it + forgot to flag it for you so thanks for catching it.

PR #1020, opened 2022-05-10
Vitess OnlineDDL no longer experimental

shlomi-noach:
We consider Vitess's Online DDL to be production grade. This includein particular includes `vitess` strategy migrations, based on VReplication.

This PR updates the docs to remove "experimental" notes, and moreover indicate that Online DDL is the preferred way of running schema migrations in Vitess.


rohit-nayak-ps (approved):
lgtm

deepthi (approved):
Are we intentional about marking pt-osc as stable too?

shlomi-noach:
@deepthi good question about `pt-osc`. I don't have a good answer as I don't have any information about usage, I'm good to keep it as experimental.

shlomi-noach:
The docs now clarify `pt-osc` is experimental. If anyone is wondering, this is because we have not had any feedback from the community with regard `pt-osc` use in `vitess` online schema changes.

PR #1025, opened 2022-05-12
Update Reserved Conn documentation

Anders-PlanetScale:
Added a line to Reserved Conn feature description that mentions the vtgate flag `enable_system_settings`.

I updated v13 and v14 docs. v12 docs don't mention this var at all, so I did not want to add a line mentioning it without also updating the rest of the documentation to follow. I assume v12 documentation is of less importance at this time.

This was done to clearly call out, early in the Reserved Conn documentation what variable is used to enable/disable reserved conns. I reviewed the rest of the documentation and found it to be robust and clear as long as all links are followed and the documentation read in its entirety.


hallaroo (approved):
LG;TM

PR #1029, opened 2022-05-16
Correct reshard -skip_schema_flag explanation

msolters:
Logic in the explanation of the flag was backwards to what the flag actually does.

Also some whitespace fixes.


mattlord (approved):
Thanks!

rohit-nayak-ps (approved):
lgtm

PR #1037, opened 2022-05-31
Disable typographer so that 2 dashes are NOT converted to em dash

mattlord:
Disable the goldmark renderer's typographer extension: https://github.com/yuin/goldmark/#typographer-extension

Otherwise our flags with double dashes (which are all of them in v14+) — when not in code blocks or other special formatting — still look like single dashes as it's still a single dash character in front of the flag name, it's just the elongated em/en dash character.

For comparison:
  - typographer enabled: https://vitess.io/docs/14.0/reference/programs/vtctld/
  - typographer disabled: https://deploy-preview-1037--vitess.netlify.app/docs/14.0/reference/programs/vtctld/

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -127,6 +127,9 @@ icon = "magic"
 [markup.goldmark.renderer]
 unsafe = true
 
+[markup.goldmark.extensions]
+typographer = false
+
 [mediaTypes]
 "text/netlify" = { }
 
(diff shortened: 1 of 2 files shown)


deepthi (approved):
The ideal fix for this would be to change the reference pages and remove the dashes altogether. There's no reason why they are needed in a list of the flags.
For now let's merge this because otherwise the reference pages are misleading. If it causes other problems we can fix forward.
Thank you for taking care of this.

PR #1046, opened 2022-06-07
Update flag related syntax in Reference section

### LINKED ISSUE #1042: Flag syntax needs to be changed in reference docs
As documented in [VEP-4](https://github.com/vitessio/enhancements/blob/main/veps/vep-4.md), we have started the process of migrating to standard `pflag` usage.
Some of the docs don't yet reflect the recommended usage, so we need to fix those
https://vitess.io/docs/14.0/reference/programs/vtexplain/
https://vitess.io/docs/14.0/reference/programs/vtgate/

The rest (mysqlctl, vtctl, vtctld, vtadmin, vttablet) are correct. However, there is inconsistency in how the flags are listed. In some cases they are prefixed with `--`, in others, there is no prefix. We should make them all standard, my suggestion is to drop the `--` prefix.

mattlord:
This audits the website pages underneath [the main `Reference` left-hand nav section](https://vitess.io/docs/14.0/reference/) to apply the new CLI flag and parameter rules for v14.

Fixes: https://github.com/vitessio/website/issues/1042

After this PR there's just the Contribute, Design Docs, Trouble Shoot, and FAQ sections left in [the main left-hand nav](https://vitess.io/docs/14.0/).

And it looks like [the `Design Docs` section](https://vitess.io/docs/14.0/design-docs/) is the only one of them that requires changes.


mattlord:
After this PR there's just the Contribute, Design Docs, Trouble Shoot, and FAQ sections left in the main left-hand nav.

And it looks like the `Design Docs` section is the only one of them that requires changes.

deepthi (commented):
mysqlctl and vtadmin need to be changed as well to be consistent. The mysqlctl page lists flags without the prefix.
vtadmin also lists flags without a prefix and one of the flags it lists is `v` which would become `-v`
I actually like the way the vtadmin reference page is laid out. One glaring omission on the other pages is that default values are not listed. Reworking all the reference pages to be similar is going to be another big task.
@ajm188 thoughts on the vtadmin flag reference?

ajm188:
> mysqlctl and vtadmin need to be changed as well to be consistent. The mysqlctl page lists flags without the prefix.
> vtadmin also lists flags without a prefix and one of the flags it lists is `v` which would become `-v`
> I actually like the way the vtadmin reference page is laid out. One glaring omission on the other pages is that default values are not listed. Reworking all the reference pages to be similar is going to be another big task.
> @ajm188 thoughts on the vtadmin flag reference?

i'm not sure what the question is, exactly. i agree with what you've laid out here

mattlord:
@deepthi I updated the `mysqlctl` page and also copied all of the changed files to the new 15.0 docs. I'm personally fine with `vtadmin` being different here as it's all new and will be a different way of interfacing with Vitess. We can IMO use that as a way to slowly establish new ways of doing/documenting things and then transition everything else over time.

deepthi (approved):

PR #1050, opened 2022-06-08
updated flags to use two hyphens for vtexplain

systay:
Signed-off-by: Andres Taylor <andres@planetscale.com>


systay:
We should just use #1046

PR #1051, opened 2022-06-08
Automate release work

frouioui:
This PR automates some of the release work that has to be done on the website repository. Two scripts are introduced: `rc_release.sh` and `ga_release.sh`. The former is used to create the next version's doc directory and change the sidebar attribute (Current/RC/Development). The latter simply changes the sidebar attribute, `RC` becomes `Current`.

### DIFF
--- a/tools/rc_release.sh
+++ b/tools/rc_release.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+next_release=$1
+
+for lang in {en,zh} ; do
+  cp -r content/${lang}/docs/$((next_release-1)).0 content/${lang}/docs/${next_release}.0
+
+  sed -i.bak 's/(Development)/(RC)/' ./content/${lang}/docs/${next_release}.0/_index.md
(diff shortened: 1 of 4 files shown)


tools/rc_release.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+next_release=$1
+
+for lang in {en,zh} ; do
+  cp -r content/${lang}/docs/$((next_release-1)).0 content/${lang}/docs/${next_release}.0
+
+  sed -i.bak 's/(Development)/(RC)/' ./content/${lang}/docs/${next_release}.0/_index.md
+  sed -i.bak 's/Under construction, development release./Release Candidate./' ./content/${lang}/docs/${next_release}.0/_index.md
```
deepthi, line 9: If we ever change the text describing the development release, we'll need to change this script as well, right?

deepthi (approved):

tools/rc_release.sh
frouioui: yes, that's correct

PR #1054, opened 2022-06-13
Add vtctldclient docs

ajm188:
This adds docs for `vtctldclient` autogenerated via `cobra/doc.GenMarkdownTreeCustom`. I'm working on generalizing the backing code, but I think me just committing this is fine for 14.0 (but when we start moving more commands over in the next release, we should 100% have this tooling integrated).

Anyway, here's the steps (from vitessio/vitess):

```
➜  vitess git:(andrew/vtctldclient/autodoc) ✗ go build -o docgen ./go/cmd/vtctldclient/docgen
➜  vitess git:(andrew/vtctldclient/autodoc) ✗ ./docgen -d ./go/cmd/vtctldclient/doc
➜  vitess git:(andrew/vtctldclient/autodoc) ✗ mv ./go/cmd/vtctldclient/doc/vtctldclient.md ./go/cmd/vtctldclient/doc/_index.md
➜  vitess git:(andrew/vtctldclient/autodoc) ✗ cp ./go/cmd/vtctldclient/doc/*.md ~/dev/vtwebsite/content/en/docs/14.0/reference/programs/vtctldclient
```

(when I publish the tooling, that'll all be in a `make` target)


GuptaManan100 (commented):
GitHub is making this impossible to review! But I ❤️ it. Do we want one file for each command though or 1 long file?

ajm188:
> GitHub is making this impossible to review!

I'd recommend just looking at the deploy preview!! Every file is "the same" (except for `vtctldclient` which contains the summary).

> Do we want one file for each command though or 1 long file?

This is the standard way that cobra auto doc works, so I'd rather stick with that than try to shoe-horn it into some other setup unless we have a specific reason to do so.

deepthi:
Clicking links on the preview is returning `404`.
E.g. https://deploy-preview-1054--vitess.netlify.app/docs/14.0/reference/programs/vtctldclient/vtctldclient_AddCellInfo.md

content/en/docs/14.0/reference/programs/vtctldclient/vtctldclient_GetBackups.md
```diff
@@ -0,0 +1,32 @@
+---
+title: GetBackups
+series: vtctldclient
+description:
+---
+## vtctldclient GetBackups
+
+
```
deepthi: Some commands like this one show an empty description on the preview. I think we need to add the right text to the source and regenerate.

ajm188: yeah, i can go through the actual commands in vitessio/vitess tomorrow and add all those

ajm188:
> Clicking links on the preview is returning `404`.
> E.g. https://deploy-preview-1054--vitess.netlify.app/docs/14.0/reference/programs/vtctldclient/vtctldclient_AddCellInfo.md

ah shoot, i see the problem, will fix it up!

content/en/docs/14.0/reference/programs/vtctldclient/vtctldclient_GetBackups.md
deepthi: will that be a follow up PR? If yes, this LGTM

ajm188: opened: https://github.com/vitessio/vitess/pull/10502. after addressing any feedback there, i'll regenerate and we can merge this one!

mattlord (commented):
This is friggin awesome! 😃 

Is `--server` really optional with `vtctldclient`? I see some commands imply that it is, e.g.
```
vtctldclient UpdateCellInfo [--root <root>] [--server-address <addr>] <cell>
```

And some commands don't mention it at all, e.g.:
```
vtctldclient DeleteKeyspace [--recursive|-r] [--force|-f] <keyspace> [flags]
```

Although for the latter I now see that it mentions those separately as being inherited. So maybe that part is fine (although I'd expect those to be in the usage output too).

ajm188:
> This is friggin awesome! 😃
> 
> Is `--server` really optional with `vtctldclient`? I see some commands imply that it is, e.g.
> 
> ```
> vtctldclient UpdateCellInfo [--root <root>] [--server-address <addr>] <cell>
> ```
> 
> And some commands don't mention it at all, e.g.:
> 
> ```
> vtctldclient DeleteKeyspace [--recursive|-r] [--force|-f] <keyspace> [flags]
> ```
> 
> Although for the latter I now see that it mentions those separately as being inherited. So maybe that part is fine (although I'd expect those to be in the usage output too).

You're mistaking the `--server-address` argument which is for _only_ `{Add,Update}CellInfo` with the global `--server` flag that's defined at the command root. 

`--server` lives [here](https://github.com/vitessio/vitess/blob/main/go/cmd/vtctldclient/command/root.go#L96) and [is required](https://github.com/vitessio/vitess/blob/main/go/cmd/vtctldclient/command/root.go#L84-L93)

ajm188:
It also does appear in each command's docs, e.g. https://deploy-preview-1054--vitess.netlify.app/docs/14.0/reference/programs/vtctldclient/vtctldclient_addcellinfo/#options-inherited-from-parent-commands

ajm188:
@deepthi okay i regenerated (and added the 15.0 versions), this should be good to go!

mattlord:
> > This is friggin awesome! 😃
> > Is `--server` really optional with `vtctldclient`? I see some commands imply that it is, e.g.
> > ```
> > vtctldclient UpdateCellInfo [--root <root>] [--server-address <addr>] <cell>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And some commands don't mention it at all, e.g.:
> > ```
> > vtctldclient DeleteKeyspace [--recursive|-r] [--force|-f] <keyspace> [flags]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Although for the latter I now see that it mentions those separately as being inherited. So maybe that part is fine (although I'd expect those to be in the usage output too).
> 
> You're mistaking the `--server-address` argument which is for _only_ `{Add,Update}CellInfo` with the global `--server` flag that's defined at the command root.

Ah, right. Sorry!


> `--server` lives [here](https://github.com/vitessio/vitess/blob/main/go/cmd/vtctldclient/command/root.go#L96) and [is required](https://github.com/vitessio/vitess/blob/main/go/cmd/vtctldclient/command/root.go#L84-L93)

It doesn't seem to be required on the command pages though, the command usage doesn't show it without square brackets to indicate it's non-optional and in the inherited section it says:
```
--server string             server to use for connection
```

I'd prefer to show those inherited things in the full usage output for each command, but can we at least add e.g. `--server string             server to use for connection (required)` as I don't see an indication that required is the default unless optional is specified either. 

It's a nit obviously, and doesn't need to hold this up (we have no docs ATM so this is obviously much better). I'll approve so that you're unblocked. Thanks!

mattlord:
One general note, not specifically related to this PR, is that we don't seem to have a uniform way of showing usage related info when you must specify 1 of N options. For example, in some cases we use ~ this to signify that one of the two is required:
```
vtctldclient UpdateCellInfo {--root <root> | --server-address <addr>} <cell>
```

We can take the lead from pflags on that and update the docs everywhere to match over time.

mattlord (approved):

ajm188:
lemme fix that real quick!

as for the standardization, totally agree. it's grown organically over time, and before v15 release we should sit down (metaphorically 😂 ) and articulate what conventions we want to stick to (and then fix all my sins)

deepthi (approved):
LGTM

ajm188:
piled on the `--server` change in https://github.com/vitessio/website/pull/1055 (specifically https://github.com/vitessio/website/pull/1055/commits/c9eff41ab4a4ac3d0a81ae2e5df54fc939895334)

PR #1058, opened 2022-06-15
Fix section page filtering partials

### LINKED ISSUE #1057: Contribute pages are organized incorrectly
This seems to be a problem in all versions of the docs.

Clicking on `Contribute` on the main docs page takes you to 
https://vitess.io/docs/14.0/contributing/

which then has
Screen Shot 2022-06-15 at 10 34 50 AM

References is a hyperlink to an empty page at https://vitess.io/docs/14.0/contributing/reference/
Screen Shot 2022-06-15 at 10 43 31 AM

I need to go back and look at what this looked like earlier to be sure but I think that `References` should contain links to `Parser and AST`/`Release Notes`/`Good First Issue` and the `Contribute` page should duplicate everything from the main docs page under `Contribute`.
What this means right now is that the GitHub workflow page is no longer linked from the main `Contribute` page and that is not good.

ajm188:
This fixes #1057.

The `.Pages` => `.RegularPages` was a bad copy-paste from me in #1054, but doesn't solve the root problem. Apparently "not in" does not work as advertised, but doing two rounds of filtering does 🤷.

I checked a few other Section pages that should have these listings and they looked correct to me.

### DIFF
--- a/layouts/partials/docs/pages-in-section.html
+++ b/layouts/partials/docs/pages-in-section.html
@@ -1,5 +1,5 @@
-{{ $skipSections := slice "vtctl" "vtctldclient" }}
-{{ $pages := where .RegularPages "Params.series" "not in" $skipSections }}
+{{ $pages := where .Pages "Params.series" "!=" "vtctl" }}
+{{ $pages := where $pages "Params.series" "!=" "vtctldclient" }}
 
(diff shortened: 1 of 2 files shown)


deepthi (approved):
👍

PR #1059, opened 2022-06-15
Update section layout to include content for nested sections

ajm188:


deepthi:
I don't think this flows well
https://deploy-preview-1059--vitess.netlify.app/docs/15.0/contributing/
The content shows up after all the links / list of pages.
Even if we changed the layout to put the content above the links, I worry that then the links are less obvious.
Best thing might be to close this PR and move the content to a separate page (Getting Started?). Then `_index.md` can be empty and people can follow the link to `Getting Started`

ajm188:
Sure, that works for me!!

deepthi:
Ultimately we decided to re-org the pages. 
- Removed `Reference` 
- Moved all the pages in the `Contribute` section to the same level
- Also show `Content` for sections which have it
- Reorder pages using weights so that they are in the order in which people should read them

These changes have been made only in 14.0 and 15.0, not in older versions.

deepthi (approved):

Issue #977, opened 2022-02-16
Website search usability issues


mattlord:
The current website search (Algolia) is of limited practical use today for the following reasons:
1. The results include no information to identify which specific page it would take you to (e.g. reference or user guide for vtgate)
2. We now have versioned docs so the same page is repeated N times in the results (and because of issue 1, you cannot see which version a result is for). When combined with the fixed number of max results shown, you often cannot get to what you want

We should see if we can add any sort of preview or metadata to the search results so that you can see where it will take you w/o having  to click on it. 

We should see if we can default to ONLY searching the [`$current` version](https://github.com/vitessio/website/blob/098e06a996c4359bd79b321f07f2966a498e1acf/config.toml#L16) of the docs.

nate-double-u:
I opened two PRs on etcd.io's website repo a while ago relating to search:
* https://github.com/etcd-io/website/pull/403
* https://github.com/etcd-io/website/pull/392

Their site is set up a little differently to this one, but Lunr or Google may be a good solution to your search issues. You can see how the search works on the https://etcd.io site. Since vitess.io isn't using docsy it may be different to set up, but this may be a good starting point.

chalin:
@mattlord, would a switch to Lunr be an acceptable solution?

mattlord:
Hi @chalin! 

Yeah, I don't think we have any real preferences on the implementation used for vitess.io. Any pointers on how we'd do that? 

I guess we'd glean that from looking at the two noted examples where etcd did this?
* https://github.com/etcd-io/website/pull/403
* https://github.com/etcd-io/website/pull/392

Are there any other suggestions, requirements, ideas on the CNCF side that we should know about?

Thanks again!

chalin:
Hi @mattlord - no, nothing else comes to mind. If the Vitess website was using the [Docsy](https://github.com/google/docsy) theme, then the switch to Lunr would be simple and quick to do. While that isn't the case, you can always look at the Docsy implementation for inspiration. (I'd be glad to lend a hand, but it looks like it wouldn't be before the start of 22Q3.)

derekperkins:
Who controls our Algolia account? Is it @lucperkins or the CNCF? It shouldn't be too hard to configure Algolia to do that correctly.

deepthi:
@chalin @nate-double-u do you know the answer to the algolia question?

nate-double-u:
[...]

PR #1070, opened 2022-06-24
Change algolia account to Vitess-owned account and filter search results by version of docs currently being looked at

llibdude:
related: #https://github.com/vitessio/website/issues/977

**Note**
I am still waiting to hear back from Algolia support on if we need to sign up [here](https://docsearch.algolia.com/apply/) for access to v3 of if since we had a hosted config before we were automatically upgraded and just didnt know about it. I will need to reach out again next week it looks like.

The solution in this PR is a stop-gap or alternative. It is sitting in `Draft` for now so that we can discuss it.

**Description**

This PR changes our algolia config from one managed by Algolia to https://github.com/algolia/docsearch-configs/blob/master/configs/vitess.json to an Algolia account owned by Vitess. Note that this **is not** an upgrade from v2 to v3 of docsearch, this solution is still using v2.

In addition, a Github action has been added to crawl the Vitess docs and generate the Algolia records.

_Versioning_
This solution supports searching the docs in any version.  Since the docs are set up where you can look at any version on the left tree panel, a version dropdown next to search didn't seem appropriate (perhaps we can change this in a future redesign).  Instead, your search results will filter based on what docs you are currently looking at, and if you are on the homepage it will use the `current` version (v14).  Here is a video example - keep in mind this is not an exhaustive list of search results as we would need to pay money to store all the records (10k limit for free account).

https://user-images.githubusercontent.com/5255924/175724267-b9658458-003a-4f67-b305-99afd4c6ee27.mov

The video starts with no version of docs selected and shows the current version (v14) results. Then I select v11 docs and search, it shows v11 results. I do that to all versions proving it's working.

**Things left to do before merging**
1. Set up the Algolia secrets in github for this repo (see the github action)
2. Upgrade the account ($$$) to handle more records since we are exceeding them.

**Things to do in the future if we go this route:**
1. Dynamically manage the Vitess versions in `vitess_config.json` so we do not need to manage a hard coded list. Here is a potential example: https://github.com/algolia/docsearch-configs/blob/75a4d86bf367d841bc4e67b2f4265be5bdbb3b59/configs/cordova.json#L9
2. Support docs not only for english but our other languages (Chinese)


deepthi:
Thanks @llibdude. Let me know what you hear back from Algolia support.

nate-double-u:
Has Algolia gotten back? It looks like it took about a day to exhaust the records we can use on the free tier.

deepthi:
superseded by #1119

PR #1087, opened 2022-07-05
Remove blank `description` from all vtctldclient doc frontmatter

ajm188:
This doesn't actually do anything (I thought I needed it when I first wrote the codegen), and I've removed it from the template in https://github.com/vitessio/vitess/pull/10635

### DIFF
--- a/content/en/docs/15.0/reference/programs/vtctldclient/_index.md
+++ b/content/en/docs/15.0/reference/programs/vtctldclient/_index.md
@@ -1,7 +1,6 @@
 ---
 title: vtctldclient
 series: vtctldclient
-description:
 ---
 ## vtctldclient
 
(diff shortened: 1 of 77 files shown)


deepthi:
`description` shows up as a sub-title. Example: https://vitess.io/docs/15.0/reference/vreplication/create/
Can you confirm that you actually don't want a sub-title and hence it is good to remove the `description` field?

ajm188:
> Can you confirm that you actually don't want a sub-title and hence it is good to remove the `description` field?

Yep, when I was first writing the autodoc tool I was struggling with something else (can't remember) and saw some other pages had blank descriptions so I added it during debugging and never removed it. It was an oversight to begin with.

ajm188:
There are a bunch of them, for reference:

```
git grep --name-only "description:$"  -- **/*.md | wc -l 
     135
```

deepthi (approved):

PR #1088, opened 2022-07-06
schema tracking doc improvement

harshit-gangal:
Highlighting more benefits of enabling schema tracking

### DIFF
--- a/content/en/docs/14.0/reference/features/schema-tracking.md
+++ b/content/en/docs/14.0/reference/features/schema-tracking.md
@@ -8,9 +8,9 @@ VTGate does not natively track table schema. Users are allowed to provide an aut
 
 The schema tracking functionality alleviates this issue and enable VTGate to plan more queries. When using schema tracking, VTGate keeps an authoritative list of columns on all tables. The following query set can be planned:
 
-* `SELECT *` cross-shard queries that need evaluation at the VTGate level.
-* Queries that are not able to resolve columns dependencies. For instance: queries with no table qualifier in the projection/filter list.
-* Evaluation improvement in Aggregations, Group By, Having, Limit, etc. clauses that require processing of records at VTGate level.
+1. `SELECT *` cross-shard queries that need evaluation at the VTGate level.
+2. Queries that are not able to resolve columns dependencies. For instance: queries with no table qualifier in the projection/filter list.
+3. Evaluation improvement in Aggregations, Group By, Having, Limit, etc. clauses that require processing of records at VTGate level. VTGate will not require `weight_string()` value for the evaluation and can compare the values directly. 
 
 ## VTGate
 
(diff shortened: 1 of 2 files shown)


content/en/docs/14.0/reference/features/schema-tracking.md
```diff
@@ -10,6 +10,7 @@ The schema tracking functionality alleviates this issue and enable VTGate to pla
 
 * `SELECT *` cross-shard queries that need evaluation at the VTGate level.
 * Queries that are not able to resolve columns dependencies. For instance: queries with no table qualifier in the projection/filter list.
+* Evaluation improvement in Aggregations, Group By, Having, Limit, etc. clauses that require processing of records at VTGate level.
```
deepthi:
```suggestion
* Expressions in Aggregations, Group By, Having, Limit, etc. clauses that require processing of records at VTGate level.
```

deepthi: How does this change sound? Is this accurate?

harshit-gangal: when schema tracking is on we do not rely on `weight_string` and use the collation information for comparison.
that is why mentioned it as improvement.

frouioui: I think to go with the paragraph above: `... The following query set can be planned:`, we should change the beginning of this list item by something like: `* Queries with any of: Aggregations, Group By, Having, Limit, etc. clauses ...`

deepthi: @harshit-gangal my question was on my suggested text, not on your original text.

harshit-gangal: without schema tracking also we are able to evaluate expressions at vtgate, with schema tracking we have collation information so we do not need additional information from mysql to help in evaluation.
For this reason I wrote `Evaluation improvement` over `Expressions` 
So, I will resolve this review.

harshit-gangal:
@deepthi need an approval to merge

content/en/docs/14.0/reference/features/schema-tracking.md
deepthi: What is the improvement? We should write down exactly what the improvement is. Performance improvement?

content/en/docs/14.0/reference/features/schema-tracking.md
```diff
@@ -8,8 +8,9 @@ VTGate does not natively track table schema. Users are allowed to provide an aut
 
 The schema tracking functionality alleviates this issue and enable VTGate to plan more queries. When using schema tracking, VTGate keeps an authoritative list of columns on all tables. The following query set can be planned:
 
-* `SELECT *` cross-shard queries that need evaluation at the VTGate level.
-* Queries that are not able to resolve columns dependencies. For instance: queries with no table qualifier in the projection/filter list.
+1. `SELECT *` cross-shard queries that need evaluation at the VTGate level.
+2. Queries that are not able to resolve columns dependencies. For instance: queries with no table qualifier in the projection/filter list.
```
deepthi:
```suggestion
2. Queries which reference columns without a table qualifier as long as the column names are unique across the tables involved in the query.
```

content/en/docs/14.0/reference/features/schema-tracking.md
harshit-gangal: done

deepthi (approved):

PR #1097, opened 2022-07-16
Add VTOrc bug documentation to 14.0

GuptaManan100:
## Description

This PR adds the VTOrc bug documentation for https://github.com/vitessio/vitess/issues/10650 to 14.0 docs. It also adds the additional steps that the user will have to take in the VTop example to get VTOrc to work.


frouioui (approved):
That looks good to me.

Issue #1106, opened 2022-07-19
Drop yarn and the lock file, and use just NPM


chalin:
I'd suggest dropping yarn and the yarn lock file and simply using NPM. Might you be open to such a simplification?

- [x] Drop yarn lock file from version control - #1107
- [x] Simplify tooling. Yarn is no longer required.

/cc @nate-double-u

PR #1117, opened 2022-07-27
made changes to navbar, footer

reveurguy:
Signed-off-by: reveurguy <59417546+reveurguy@users.noreply.github.com>

- Added background to hero,footer. 
- made changes to navbar style
- updated fontawesome version


GuptaManan100:
@reveurguy If you were doing the open-close draft-ready for getting the CI to run, then that is not required. For external contributors, netlify requires approval from a member to run the preview. I just approved it, so they'll run now.

GuptaManan100 (commented):
I like the changes but I am not sure why we have removed the `yarn.lock` file. I'll let someone with more web experience have the final word.

reveurguy:
> @reveurguy If you were doing the open-close draft-ready for getting the CI to run, then that is not required. For external contributors, netlify requires approval from a member to run the preview. I just approved it, so they'll run now.

ok, yeah i was doing that for that only😅

reveurguy:
> I like the changes but I am not sure why we have removed the `yarn.lock` file. I'll let someone with more web experience have the final word.

that was updated automatically, when i ran the site locally. it is not deleted but most of it's content was removed.

deepthi:
@reveurguy can you attach before and after screenshots? Alternatively, you can link to the specific changed pages in the preview.
For instance, I can see that you have changed the background on the homepage, but I can't tell what else has changed.

deepthi:
@nate-double-u @chalin can one of you review the changes? the diff on `yarn.lock` is huge.

reveurguy:
# Before

Screenshot 2022-07-28 at 8 57 32 PM

# After
Screenshot 2022-07-28 at 8 57 18 PM

# before 
Screenshot 2022-07-28 at 8 58 49 PM

# after
Screenshot 2022-07-28 at 8 58 34 PM

reveurguy:
initially i have made these small changes only.

deepthi:
@reveurguy can you look at this on mobile? The changes to the button shapes mean that they look kind of awkward now because one of them is rounded on the left, one is a rectangle, and the third one is rounded on the right.

File

chalin:
> @nate-double-u @chalin can one of you review the changes? the diff on `yarn.lock` is huge.

My suggestion is to drop the lock file from version control:

- #1106

chalin:
I'd suggest leaving the footer the way it was. The new styling would make the text difficult to read for some. My 2 cents.

chalin:
@thisisobate - any comments about the suggested changes?

nate-double-u:
> My suggestion is to drop the lock file from version control:

+1

(i also tend to keep the `package-lock.json` file out of version control. I know they're designed to be included, but they always seemed like more trouble than they're worth to me.)

layouts/partials/footer.html
```diff
@@ -1,11 +1,11 @@
 {{ $year       := now.Year }}
 {{ $footerLogo := "img/logos/vitess-horizontal.png" | relURL }}
 {{ $social     := site.Params.social }}
-<footer class="footer has-background-black">
+<footer class="footer has-background">
   <div class="container">
     <img class="is-footer-logo" src="{{ $footerLogo }}">
 
-    <hr class="has-background-orange">
+    <hr class="has-background-black">
```
thisisobate: revert the `hr` color back to orange as the black foreground doesn't match with the background image

layouts/partials/footer.html
```diff
@@ -1,11 +1,11 @@
...
 
     <nav class="level">
```
thisisobate, line 10: The Nav color doesn't match well with the background; can barely read the text. I think you should use a text color that has good color contrast and matches the background- white is a good candidate.

thisisobate: Also, the external link's contrast isn't great. You should increase it (very light blue color might do the trick?)

thisisobate (changes requested):
+1 on the lock file as @nate-double-u and @chalin suggested.

Add a background image to the footer or leave it as it was before? either way works tbh.

@reveurguy If you want to go with the former, then you should fix the readability issues @chalin mentioned.

layouts/partials/footer.html
thisisobate: Screenshot 2022-07-30 at 04 58 58

The Nav color doesn't match well with the background; can barely read the text. I think you should use a text color that has good color contrast and matches the background- white is a good candidate.

thisisobate: Also, the external link's contrast isn't great. You should increase it (very light blue color might do the trick?)

reveurguy:
made the  required changes

deepthi (approved):
Looks like all reviews have been addressed.
Re `yarn.lock`, more changes are required to drop `yarn` completely but they can be done separately whenever someone picks up #1106 
@reveurguy welcome to Vitess and thank you for the contribution!

PR #1119, opened 2022-07-28
Feat: implement search using Lunr

### LINKED ISSUE #977: Website search usability issues

thisisobate:
**Todos**
- [x] Add search input
- [x] Style search result
- [x] Make search results view mobile responsive
- [x] Create search index using [lunr](https://lunrjs.com/)
- [x] Implement search functionality
- [x] make search query results depending on the version of the docs
- [x] Show total search count when search is enabled
- [x] Replace the Algolia search functionality in favor of Lunr
- [x] Show versions in results pane
- [x] Dynamic positioning of the search popover using popperjs
- [x] clean up code

fixes #977
relates to https://github.com/cncf/techdocs/issues/101


deepthi:
cc @nate-double-u @chalin @llibdude

nate-double-u:
@deepthi, @thisisobate and I are looking at Lunr as an alternative to Algolia (Algoila seems to have some issues around the number of records we can use). Lunr may be a good, free alternative. But there's some exploration to do still :)

llibdude:
Hi, sorry missed this.  We are very very slowly communicating with Algolia about getting on their new docsearch project.  I believe it will give us what we need but it is good to have alternatives.

thisisobate:
Yeah...I totally understand @llibdude 
The idea behind this draft PR is to show other free alternatives we can adopt- in this case, [Lunr.js](https://lunrjs.com/).

deepthi:
I took at look at the preview. One thing that is better about this implementation is that it lists all results.
Example:
Screen Shot 2022-09-27 at 5 41 19 PM

Things that need fixing/improvement:
- The preview is unreadable because of the font color
- The first result is still somehow the oldest (v11.0). What can we do so that newer release docs show up higher?

thisisobate:
Thanks for the feedback! @deepthi 
I'm currently working on fixing all of the issues related to this PR.
Will do well to open it up for review once I'm done

thisisobate:
> The first result is still somehow the oldest (v11.0). What can we do so that newer release docs show up higher?

@deepthi The best way to avoid issues like this is to make search query results depending on the version of the docs. I already implemented the solution in my last commit.

layouts/partials/javascript.html
```diff
@@ -1,26 +1,163 @@
...
+<!-- <script type="text/javascript" src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"></script>
 <script type="text/javascript">
 docsearch({
   apiKey: '1d13fc4b241cae42000f422300df8931',
   indexName: 'vitess',
   inputSelector: '#search-bar',
   debug: false
 });
-</script>
+</script> -->
```
nate-double-u: Probably best to remove dead code entirely if we don't need it anymore, rather than having it hang around commented out.

nate-double-u (commented):
Thanks @thisisobate, this is looking good.

A couple things to consider for updates:
* Does this index all the languages? (Should it? @deepthi)
* Could we add version indicators to the results pane? It looks like we are defaulting to version 14 -- but when i search on a v15 page i get v15 results (would be good to know before clicking)
* Could we add some space above the titles in the results pane? It looks like the titles are more closely connected to the previous result than the actual result

Screen Shot 2022-09-29 at 11 33 12 AM
Screen Shot 2022-09-29 at 11 32 26 AM

nate-double-u:
It should also be noted that this removes and replaces the Algolia search functionality

thisisobate:
> Could we add version indicators to the results pane? It looks like we are defaulting to version 14 -- but when i search on a v15 page i get v15 results (would be good to know before clicking)

By default, the search returns results depending on the current selected version. Whenever you search on a v15 page, the search can only show you v15 results. I think my only confusion is: how do I identify the latest version- stable or development?

I went for stable versions cus I feel it makes more sense but I think @deepthi would have better insight on this. Your thoughts?

deepthi:
> 
> By default, the search returns results depending on the current selected version. Whenever you search on a v15 page, the search can only show you v15 results. 

This makes sense to me

> I think my only confusion is: how do I identify the latest version- stable or development?
> 
> I went for stable versions cus I feel it makes more sense but I think @deepthi would have better insight on this. Your thoughts?

When there is no version "context" searching the latest stable version is the right decision. 
If someone wants to search within the development release, then they would have to first navigate to those docs.

It will be nice to have (if possible) a version indicator on the results as @nate-double-u suggested.

thisisobate:
> It will be nice to have (if possible) a version indicator on the results as @nate-double-u suggested.

By design, search results are not meant to show which version you're looking for. You can only identify the correct version by hovering on the links. I also took some time to check out other project's websites e.g Prometheus, Thanos, etc and it's no different from my current implementation. 

Moreover, if this becomes an issue, then I would suggest we make the **development** the latest version to avoid confusions.

nate-double-u:
> Moreover, if this becomes an issue, then I would suggest we make the development the latest version to avoid confusions.

I think the last stable release should be the default if the system gets confused for whatever reason.

> By design, search results are not meant to show which version you're looking for. You can only identify the correct version by hovering on the links. I also took some time to check out other project's websites e.g Prometheus, Thanos, etc and it's no different from my current implementation.

Does the version not come in with the results? (we can see that we are making a choice of what results to show, it seems odd that we wouldn't be able to add it as a subhead to the main article title)

nate-double-u (commented):
> Moreover, if this becomes an issue, then I would suggest we make the development the latest version to avoid confusions.

I think the last stable release should be the default if the system gets confused for whatever reason.

> By design, search results are not meant to show which version you're looking for. You can only identify the correct version by hovering on the links. I also took some time to check out other project's websites e.g Prometheus, Thanos, etc and it's no different from my current implementation.

Does the version not come in with the results? (we can see that we are making a choice of what results to show, it seems odd that we wouldn't be able to add it as a subhead to the main article title)

deepthi:
> Does this index all the languages? (Should it? @deepthi)

Ideally, searching on an en page should show only en results and on zh pages should show only zh results.
I'm assuming it already works this way :)
If the question was whether search should be enabled on the Chinese pages, yes, it should.

deepthi (approved):
Looks good from my perspective. Once the version question from @nate-double-u is resolved to the satisfaction of both parties, we can merge this.

nate-double-u (approved):
Thanks for those updates @thisisobate!
/LGTM

I think we can do the language stuff in a follow up PR

thisisobate:
cc: @deepthi

deepthi:
Thank you for this work @thisisobate!

nate-double-u:
> Thank you for this work @thisisobate!

+1, thanks @thisisobate, great work!

PR #1125, opened 2022-07-29
Fix truncated table entries on vtctl index pages

### LINKED ISSUE #1104: Truncated table entries in the vtctl page
- Visit https://vitess.io/docs/15.0/reference/programs/vtctl/
- Search for a backtick (``` ` ```), and you'll find truncated entries like the following ():
  > Screen Shot 2022-07-19 at 14 19 21

  This arises because there are bar characters (`|`) inside code-formatted text,
  like this ``` `some code text containing a | char` ```.

  **Hugo 0.82.0 and later requires bar characters in such situations to be escaped**.

Related:

- #1004

/cc @nate-double-u

ajm188:
Closes #1104

cc @chalin


GuptaManan100 (approved):
Looks good to me

PR #1127, opened 2022-08-03
using make command (shortcut) when running the docker image

mahmoodahmad100:
This PR updates the docs for "Local Install via Docker" (Run the docker image section) in which we use the shortcut command to run the docker image


deepthi (approved):
Makes sense. Thank you for the contribution!

mahmoodahmad100:
> Makes sense. Thank you for the contribution!

@deepthi it's my pleasure, thanks a lot!

PR #1128, opened 2022-08-03
Fix resharding command in documentation

EmadMokhtar:
The command has a wrong parameter. If you run it you'll get this error

```
vitess@c05574c06193:/vt/local$ vtctlclient InitShardPrimary -force customer/80- zone1-400
unknown shorthand flag: 'f' in -force
```


GuptaManan100 (approved):
Good catch.

GuptaManan100:
@EmadMokhtar Could you check if the same change is also required in v15 docs?

EmadMokhtar:
> @EmadMokhtar Could you check if the same change is also required in v15 docs?

Pushed the fix to 1.15 as well.

GuptaManan100 (approved):
Thank-you for your contribution. I'll merge as soon as the tests pass

PR #1130, opened 2022-08-05
source keyspace is commerce not customer (MoveTables -> Reverse workflow section)

mahmoodahmad100:
this PR puts the right keyspace in MoveTables page (User Guides -> Migrations -> MoveTables -> Reverse workflow), it should be commerce as it's the source keyspace

### DIFF
--- a/content/en/docs/11.0/user-guides/migration/move-tables.md
+++ b/content/en/docs/11.0/user-guides/migration/move-tables.md
@@ -399,7 +399,7 @@ $ vtctlclient GetRoutingRules commerce
 
 ## Reverse workflow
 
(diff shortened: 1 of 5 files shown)


mattlord (commented):
Thank you, @mahmoodahmad100 ! Since we're editing them, do you mind adding a second dash to the `--reverse_replication` flag for the 14.0 and 15.0 pages? From v14 on all long flag names should use double dashes (as we're moving to spf13 cobra and viper for config management). That also slipped through the cracks.

mahmoodahmad100:
> Thank you, @mahmoodahmad100 ! Since we're editing them, do you mind adding a second dash to the `--reverse_replication` flag for the 14.0 and 15.0 pages? From v14 on all long flag names should use double dashes (as we're moving to spf13 cobra and viper for config management). That also slipped through the cracks.

@mattlord thanks for letting me know that, I have added the second dash, thanks!

mattlord (approved):
:heart: Thanks again!

mahmoodahmad100:
@mattlord it's my pleasure, thanks a lot!

PR #1132, opened 2022-08-11
Drop yarn requirement from toolchain

### LINKED ISSUE #1106: Drop yarn and the lock file, and use just NPM

chalin:
- Closes #1106

/cc @deepthi @doeg

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,19 +6,12 @@ This repo houses the assets used to build the website at https://vitess.io.
 
 ## Running the site locally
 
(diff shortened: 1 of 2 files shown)


nate-double-u (approved):
/lgtm

PR #1136, opened 2022-08-18
Documenting vreplication flag: vreplication_max_time_to_retry_on_error

shlomi-noach:
This PR documents the flag `--vreplication_max_time_to_retry_on_error`, introduced in https://github.com/vitessio/vitess/pull/10429. The default for this flag is modified in https://github.com/vitessio/vitess/pull/11031. https://github.com/vitessio/vitess/pull/11031 is expected to be backported to v14, and so this PR updates both `v15` and `v14`.

Related:
- https://github.com/vitessio/vitess/pull/11031
- https://github.com/vitessio/vitess/issues/10783

### DIFF
--- a/content/en/docs/15.0/reference/vreplication/flags.md
+++ b/content/en/docs/15.0/reference/vreplication/flags.md
@@ -128,7 +128,7 @@ stalled streams after _vreplication_retry_delay_ seconds
 #### vreplication_max_time_to_retry_on_error
 
 **Type** duration\
-**Default** 0\
+**Default** 0 (unlimited)\
 **Applicable on** target
 
 Stop automatically retrying when we've had consecutive failures with the same error for this long after the first occurrence (default 0, meaning no time limit).
--- a/content/en/docs/15.0/reference/vreplication/flags.md
+++ b/content/en/docs/15.0/reference/vreplication/flags.md
@@ -125,6 +125,14 @@ When enabled, vttablet will start the _tracker_ which runs a separate vstream th
 The target might encounter connection failures during a workflow. VReplication automatically retries
 stalled streams after _vreplication_retry_delay_ seconds
 
+#### vreplication_max_time_to_retry_on_error
+
+**Type** duration\
+**Default** 0\
+**Applicable on** target
+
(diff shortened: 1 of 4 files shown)


content/en/docs/15.0/reference/vreplication/flags.md
```diff
@@ -125,6 +125,14 @@ When enabled, vttablet will start the _tracker_ which runs a separate vstream th
 The target might encounter connection failures during a workflow. VReplication automatically retries
 stalled streams after _vreplication_retry_delay_ seconds
 
+#### vreplication_max_time_to_retry_on_error
+
+**Type** duration\
+**Default** 0\
```
mattlord: IMO, we could clarify here:
```
**Default** 0 (unlimited)\
```

mattlord (approved):
Have one minor suggestion and a more important request (see subsequent comment), but looks good!

mattlord:
@shlomi-noach one other not-required request here... I think it would be a good idea to add documentation on that flag here for v14 and v15: https://vitess.io/docs/15.0/reference/vreplication/flags/

Then we can expound on it too like what if any related status/metrics exist, when you may want to change the default and why, etc. 

If you don't want to do that then let me know and I'll create a ToDo for it.

mattlord (changes requested):
This `--vreplication_max_time_to_retry_on_error` flag should also be listed on the `vttablet` flag reference pages:
  - https://vitess.io/docs/14.0/reference/programs/vttablet/
  - https://vitess.io/docs/15.0/reference/programs/vttablet/

Those pages should be a complete list of `vttablet` flags to matches the help output. Unfortunately that's not true today (e.g. some OnlineDDL related ones aren't there) but we should try our best to keep that in sync with the full flag set and help output.

Thank you! If you want, I'm more than happy to push commits to this PR for that part.

content/en/docs/15.0/reference/vreplication/flags.md
shlomi-noach: done

shlomi-noach:
> This --vreplication_max_time_to_retry_on_error flag should also be listed on the vttablet flag reference pages:

Done!

mattlord (approved):
❤️

PR #1139, opened 2022-08-23
Adds warning stating that MariaDB 10.2 will not be guaranteed support in future patches

GuptaManan100:
## Description

Documentation for https://github.com/vitessio/vitess/pull/11074 and https://github.com/vitessio/vitess/pull/11073

### DIFF
--- a/content/en/docs/12.0/overview/supported-databases.md
+++ b/content/en/docs/12.0/overview/supported-databases.md
@@ -24,6 +24,8 @@ Vitess supports the core features of MariaDB versions 10.0 to 10.3. Vitess [does
 
 {{< warning >}}MariaDB support has been EOL'd in Vitess 14.0.{{< /warning >}}
 
+{{< warning >}}MariaDB 10.2 EOL'd on 23 May, 2022 and Vitess patch releases after 12.0.5 will not guarantee support for it. {{< /warning >}}
+
 ## See also
 
 + [MySQL Compatibility](../../reference/compatibility/mysql-compatibility/)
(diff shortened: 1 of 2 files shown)


content/en/docs/12.0/overview/supported-databases.md
```diff
@@ -24,6 +24,8 @@ Vitess supports the core features of MariaDB versions 10.0 to 10.3. Vitess [does
 
 {{< warning >}}MariaDB support has been EOL'd in Vitess 14.0.{{< /warning >}}
 
+{{< warning >}}MariaDB 10.2 EOL'd on 23 May, 2022 and Vitess patch releases after 12.0.5 will not guarantee support for it. {{< /warning >}}
```
mattlord:
```
{{< warning >}}[MariaDB 10.2 was EOL'd on May 23, 2022](https://mariadb.org/about/#maintenance-policy) and thus Vitess support for it is also EOL in 12.0.5 and later.{{< /warning >}}
```

mattlord (commented):
I think we can simplify the note and we should add a link. Otherwise it's spot on.

content/en/docs/12.0/overview/supported-databases.md
GuptaManan100: This is much better 🙇‍♂️

GuptaManan100: Fixed

mattlord (approved):
Thanks!

rsajwani (approved):

PR #1145, opened 2022-09-01
doc: enabling settings pool on vttablet

harshit-gangal:
This PR documents the new settings pool flag on vttablet.

### DIFF
--- a/content/en/docs/15.0/reference/features/connection-pools.md
+++ b/content/en/docs/15.0/reference/features/connection-pools.md
@@ -109,23 +109,30 @@ As a result, connection pools should be sized mindful of the capacity of the und
 
 ## Other relevant pool-related variables
 
-### vttablet user limit flag:
+### vttablet user limit
+Flag: `--transaction_limit_per_user` 
 
-#### `--transaction_limit_per_user` 
-            
  * (default 0.4)
  * This flag determines the fraction of connections in the **transaction_pool** and **found_rows_pool** that can be used by a single user. 
  The username is passed to vttablet from vtgate.  
  If you are using a limited set of users, you may want to increase this limit.  
  Or disable this limit feature by setting `--transaction_limit_by_username` to `false` as the default is `true`.
  This option only comes into play if the TX limiter is enabled by `--enable_transaction_limit`, which it is not by default.
- 
- ## `--enable_system_settings`
 
-This vtgate flag converts pool connections into [reserved/dedicated session connections](../../query-serving/reserved-conn/#enabling-reserved-connections) that live for the life of the vtgate session.  The pool is then refilled.  Thus their lifecycle is outside of that of the usual pool
-connections. As a result, the number of MySQL server connections used by
-vttablet may be significantly higher than expected from the pool settings if
-you have `--enable_system_settings` enabled.
+### vtgate system settings
+Flag: `--enable_system_settings`
+
+This vtgate flag enables the user to modify a [subset of system settings](https://github.com/vitessio/vitess/blob/main/go/vt/sysvars/sysvars.go#L166-L211) on the MySQL. This is done using the mechanism of [reserved connection](../../query-serving/reserved-conn/#enabling-reserved-connections).
+Once a reserved connection is created, it lives for the life of the vtgate session. These connections lives outside of the connection pool and as a result, the number of MySQL server connections used by
+vttablet may become significantly higher than expected connection based on the pool settings.
+
+### vttablet settings pool
+Flag: `--queryserver-enable-settings-pool`
+
+This vttablet flag enables pooling of connections with modified settings.
+This overcomes the issue described in [vtgate system settings](#vtgate-system-settings).
+Both of these should be enabled to keep the feature of modifying system settings without foregoing the benefits of connection pooling.
+
 
 ## Calculating maximum db connections used by vttablet
 
--- a/content/en/docs/15.0/reference/features/connection-pools.md
+++ b/content/en/docs/15.0/reference/features/connection-pools.md
@@ -124,7 +124,7 @@ Flag: `--enable_system_settings`
 
 This vtgate flag allows clients to modify a [subset of system settings](https://github.com/vitessio/vitess/blob/main/go/vt/sysvars/sysvars.go#L166-L211) on the MySQL. This is done using the mechanism of [reserved connection](../../query-serving/reserved-conn/#enabling-reserved-connections).
 Once a reserved connection is created, it lives for the life of the vtgate client session. These connections live outside of the regular connection pool and as a result, the number of MySQL server connections used by
-vttablet may become significantly higher than expected connection based on the pool settings.
+vttablet may become significantly higher than what you might expect based on the pool settings.
 
 ### vttablet settings pool
 Flag: `--queryserver-enable-settings-pool`
--- a/content/en/docs/15.0/reference/features/connection-pools.md
+++ b/content/en/docs/15.0/reference/features/connection-pools.md
@@ -130,7 +130,7 @@ vttablet may become significantly higher than what you might expect based on the
 Flag: `--queryserver-enable-settings-pool`
 
 This vttablet flag enables pooling of connections with modified settings.
-This overcomes the issue described in [vtgate system settings](#vtgate-system-settings).
+This overcomes the issue described with the number of MySQL connections in [vtgate system settings](#vtgate-system-settings).
 Both of these should be enabled to keep the feature of modifying system settings without foregoing the benefits of connection pooling.
 
 
--- a/content/en/docs/15.0/reference/features/connection-pools.md
+++ b/content/en/docs/15.0/reference/features/connection-pools.md
@@ -122,7 +122,7 @@ Flag: `--transaction_limit_per_user`
 ### vtgate system settings
 Flag: `--enable_system_settings`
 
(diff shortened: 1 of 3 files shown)


deepthi (commented):
I recommend renaming the doc to "System Settings" or something like that.
"Reserved Connections" is really an implementation detail.
The docs should explain both ways of using session variables - with and without reserved connections.
Since we went into so much detail of when reserved conns are used, we have to do the same thing with settings pool.

harshit-gangal:
This document contains information about when reserved connections are used 
1. system settings
2. temporary tables
3. advisory lock

If you could propose any better name that can say all three in the title I will update it.

harshit-gangal:
We have a separate page for set statements https://vitess.io/docs/15.0/reference/query-serving/set-stmt/ 
which is more or less also has the design as well.

harshit-gangal:
This document is linked through connection pooling, I can add some additional details in `connection-pools.md` about what it means, but I do not see a better way to name this document.

content/en/docs/15.0/reference/features/connection-pools.md
```diff
@@ -109,23 +109,30 @@ As a result, connection pools should be sized mindful of the capacity of the und
...
- ## `--enable_system_settings`
 
-This vtgate flag converts pool connections into [reserved/dedicated session connections](../../query-serving/reserved-conn/#enabling-reserved-connections) that live for the life of the vtgate session.  The pool is then refilled.  Thus their lifecycle is outside of that of the usual pool
-connections. As a result, the number of MySQL server connections used by
-vttablet may be significantly higher than expected from the pool settings if
-you have `--enable_system_settings` enabled.
+### vtgate system settings
+Flag: `--enable_system_settings`
+
+This vtgate flag enables the user to modify a [subset of system settings](https://github.com/vitessio/vitess/blob/main/go/vt/sysvars/sysvars.go#L166-L211) on the MySQL. This is done using the mechanism of [reserved connection](../../query-serving/reserved-conn/#enabling-reserved-connections).
```
GuptaManan100: Shouldn't we mention that reserved connections won't be used if the lower flag is specified

harshit-gangal: This is already written

```
### vttablet settings pool
Flag: `--queryserver-enable-settings-pool`

This vttablet flag enables pooling of connections with modified settings.
This overcomes the issue described in [vtgate system settings](#vtgate-system-settings).
Both of these should be enabled to keep the feature of modifying system settings without foregoing the benefits of connection pooling.
```

deepthi:
> This document contains information about when reserved connections are used
> 
> 1. system settings
> 2. temporary tables
> 3. advisory lock
> 
> If you could propose any better name that can say all three in the title I will update it.

OK, this is fine. I will suggest some cosmetic changes to the wording and then we can get this merged.

content/en/docs/15.0/reference/features/connection-pools.md
deepthi:
```suggestion
This vtgate flag allows clients to modify a [subset of system settings](https://github.com/vitessio/vitess/blob/main/go/vt/sysvars/sysvars.go#L166-L211) on the MySQL. This is done using the mechanism of [reserved connection](../../query-serving/reserved-conn/#enabling-reserved-connections).
```

content/en/docs/15.0/reference/features/connection-pools.md
```diff
@@ -109,23 +109,30 @@ As a result, connection pools should be sized mindful of the capacity of the und
...
+Once a reserved connection is created, it lives for the life of the vtgate session. These connections lives outside of the connection pool and as a result, the number of MySQL server connections used by
```
deepthi:
```suggestion
Once a reserved connection is created, it lives for the life of the vtgate client session. These connections live outside of the regular connection pool and as a result, the number of MySQL server connections used by
```

content/en/docs/15.0/reference/features/connection-pools.md
```diff
@@ -109,23 +109,30 @@ As a result, connection pools should be sized mindful of the capacity of the und
...
+vttablet may become significantly higher than expected connection based on the pool settings.
```
deepthi:
```suggestion
vttablet may become significantly higher than what you might expect based on the pool settings.
```

content/en/docs/15.0/reference/features/connection-pools.md
```diff
@@ -109,23 +109,30 @@ As a result, connection pools should be sized mindful of the capacity of the und
...
+
+### vttablet settings pool
+Flag: `--queryserver-enable-settings-pool`
+
+This vttablet flag enables pooling of connections with modified settings.
+This overcomes the issue described in [vtgate system settings](#vtgate-system-settings).
```
deepthi:
```suggestion
This overcomes the issue described with the number of MySQL connections in [vtgate system settings](#vtgate-system-settings).
```

content/en/docs/15.0/reference/features/connection-pools.md
```diff
@@ -109,23 +109,30 @@ As a result, connection pools should be sized mindful of the capacity of the und
...
+Both of these should be enabled to keep the feature of modifying system settings without foregoing the benefits of connection pooling.
```
deepthi:
```suggestion
Both these flags should be enabled for clients to be able to modify system settings without foregoing the benefits of connection pooling.
```

content/en/docs/15.0/reference/query-serving/reserved-conn.md
```diff
@@ -31,6 +31,20 @@ MySQL connector and/or ORM sends to MySQL/`vtgate`. Or if those settings will
 result in reserved connections being employed for some/all of the application's
 sessions.
 
+### Settings pool and reserved connections
+
+There are multiple sections below that highlight how reserved connections get triggered for different use cases.
```
deepthi:
```suggestion
We will see how reserved connections get triggered for different use cases in subsequent sections of this document.
```

content/en/docs/15.0/reference/query-serving/reserved-conn.md
```diff
@@ -31,6 +31,20 @@ MySQL connector and/or ORM sends to MySQL/`vtgate`. Or if those settings will
...
+There is an issue when a reserved connection is used as it turns off the reuse of connection in the vttablet. More details about it are given [below](#number-of-vttablet---mysql-connections).
```
deepthi:
```suggestion
What we want to highlight at the beginning is that there is a known issue when a reserved connection is used as it cannot be reused by vttablet. More details about it are given [below](#number-of-vttablet---mysql-connections).
```

content/en/docs/15.0/reference/query-serving/reserved-conn.md
```diff
@@ -31,6 +31,20 @@ MySQL connector and/or ORM sends to MySQL/`vtgate`. Or if those settings will
...
+
+The current pool implementation is enhanced to keep the connections with settings in the pool and not to pin the connection to the session.
```
deepthi:
```suggestion
To solve this problem, the connection pool implementation used by vttablet has been enhanced to keep the connections with settings in the pool and not to pin the connection to the client session.
```

content/en/docs/15.0/reference/query-serving/reserved-conn.md
```diff
@@ -31,6 +31,20 @@ MySQL connector and/or ORM sends to MySQL/`vtgate`. Or if those settings will
...
+This will avoid the pitfall of reserved connection i.e. blocking MySQL connection which leads to MySQL running out of connections.
```
deepthi:
```suggestion
With this enhancement, we reduce the likelihood of MySQL running out of connections due to reserved connections, because the scenarios where we still need reserved connections are sharply reduced.
```

content/en/docs/15.0/reference/query-serving/reserved-conn.md
```diff
@@ -31,6 +31,20 @@ MySQL connector and/or ORM sends to MySQL/`vtgate`. Or if those settings will
...
+This also helps in retaining the other benefits of having a connection pool.
```
deepthi:
```suggestion
```

content/en/docs/15.0/reference/query-serving/reserved-conn.md
```diff
@@ -31,6 +31,20 @@ MySQL connector and/or ORM sends to MySQL/`vtgate`. Or if those settings will
...
+
+This feature is currently disabled by default and can be enabled with a flag `--queryserver-enable-settings-pool` on vttablet.
```
deepthi:
```suggestion
Since this is a new feature, it is currently disabled by default. It can be enabled by setting the flag `--queryserver-enable-settings-pool` on vttablet.
```

content/en/docs/15.0/reference/query-serving/reserved-conn.md
```diff
@@ -31,6 +31,20 @@ MySQL connector and/or ORM sends to MySQL/`vtgate`. Or if those settings will
...
+This only works for the cases when the system variables need a reserved connection.
+
+There are still cases like [temporary tables](#temporary-tables-and-reserved-connections) and [advisory locks](#get_lock-and-reserved-connections) where the reserved connection will continue to be used.
```
deepthi:
```suggestion
This change takes effect only for the cases when system variable changes need a reserved connection.
There are still cases like [temporary tables](#temporary-tables-and-reserved-connections) and [advisory locks](#get_lock-and-reserved-connections) where reserved connections will continue to be used.
```

deepthi (approved):
Can you make the suggested changes? Once they are done, please go ahead and merge it.
Apologies for the long review cycle on this, and thank you for your patience!

PR #1153, opened 2022-09-24
Document https://github.com/vitessio/vitess/pull/11344

mattlord:
This documents the new `--vstream-binlog-rotation-threshold ` vttablet flag and the `VStreamerFlushedBinlogs` vttablet status var added in: https://github.com/vitessio/vitess/pull/11344


rohit-nayak-ps (approved):
lgtm

PR #1167, opened 2022-10-11
Finish design doc move to unversioned sections

mattlord:
It looks like we intended to do this before but the directory was moved underneath another directory that did not have an index file so it was hidden.

You can see it now in the left hand nave here: https://deploy-preview-1167--vitess.netlify.app/docs/


deepthi (approved):
My only suggestion is to weight this section so that it is a bit lower down in the left nav.
Either after FAQ or after Resources. After Resources is what I'd prefer.

PR #1196, opened 2022-10-26
feat: document vttablet --vreplication-parallel-insert-workers

maxenglander:
Docs for https://github.com/vitessio/vitess/pull/10828.


content/en/docs/16.0/reference/vreplication/flags.md
```diff
@@ -22,6 +22,22 @@ The relay log buffers events on the target as they are received from the source.
 **relay_log_max_size** defines the maximum buffer size (in bytes). As events arrive they are stored in the relay log. The apply thread consumes these events as fast as it can. When the relay log fills up we no longer pull
 events from the source until some events are consumed. If single rows are larger than the specified buffer size, a single row is buffered at a time.
 
+#### vreplication-parallel-insert-workers
+
+**Type** integer\
+**Default** 1\
+**Applicable on** target
+
+This flag is intended as an option to improve the performance of the VReplication copy phase.
```
mattlord: We can link here: https://vitess.io/docs/design-docs/vreplication/life-of-a-stream/#full-table-copy

content/en/docs/16.0/reference/vreplication/flags.md
```diff
@@ -22,6 +22,22 @@ The relay log buffers events on the target as they are received from the source.
...
+
+During the VReplication copy phase, the target tablet reads batches of rows from the source tablet and inserts them on the target. By default, the target does this sequentially: it reads a batch, then it inserts a batch, then it reads a batch, etc. This flag adds a degree of parallelism so that, while a new batch is being read from the source, up to `--vreplication-parallel-insert-workers` may be inserting previously read batches.
```
mattlord: ... reads batches of rows in VStream packets from ...

content/en/docs/16.0/reference/vreplication/flags.md
```diff
@@ -22,6 +22,22 @@ The relay log buffers events on the target as they are received from the source.
...
+
+The performance of a VReplication stream is dependent on a number of factors, such as the hardware of the source and target tablets, the latency of the network between them, and utilization of those resources by the VReplication stream and concurrent workloads. Whether this flag improves performance depends on those factors and many others not mentioned here.
+
+A rule of thumb to follow is to see if there are idle resources (especially CPU and disk IO) on both the source and target side. If so, then increasing this flag may increase utilization of those resources, and improve copy phase performance.
```
mattlord: Any metrics to gauge effectiveness or is it all about execution time?

content/en/docs/16.0/reference/vreplication/flags.md
mattlord: I think it's worth noting that while executing the INSERTs contained in VStream packets — with each packet being processed in a single transaction — can be done in parallel, the transactions are still committed in stream order. Maybe an info note. This will head off concerns and help explain the limitations.

mattlord: This also implies that larger vstream packet sizes may help

mattlord (commented):
Looking good! We should add the new metrics here too: https://vitess.io/docs/16.0/reference/vreplication/metrics/

content/en/docs/16.0/reference/vreplication/flags.md
maxenglander: I'll add a note about effectiveness. And yeah basically boils down to execution time. When we were testing parallel inserts on MFP branches we graphed the rate of the row count metrics.

content/en/docs/16.0/reference/vreplication/flags.md
maxenglander: Strong preference to link to `#full-table-copy`, or OK to link to `#copy`? The latter feels more closely related to this flag.

content/en/docs/16.0/reference/vreplication/flags.md
maxenglander: @mattlord do you think I should include this detail about larger vstreamer packet sizes?

content/en/docs/16.0/reference/vreplication/flags.md
mattlord: Either one is good, IMO.

content/en/docs/16.0/reference/vreplication/flags.md
mattlord: I think it's worth a parenthetical, e.g.:
```
During the VReplication copy phase, the target tablet reads batches of rows in VStream packets (the size of which is managed by the [`--vstream_packet_size` flag](#vstream_packet_size)) from the source tablet and inserts them on the target. By default, the target does this sequentially: it reads a batch, then it inserts a batch, then it reads a batch, etc. This flag adds a degree of parallelism so that, while a new batch is being read from the source, up to `--vreplication-parallel-insert-workers` may be inserting previously read batches.
```

mattlord (approved):
Nice work! ❤️

PR #1197, opened 2022-10-26
DDL strategy flags: document --fast-range-rotation

shlomi-noach:
This PR documents changes made in https://github.com/vitessio/vitess/pull/10315. Namely, a new DDL strategy flag `--fast-range-rotation`. The flag was added a wihle back but was left undocumented as we were still exploring its use/viability. We have now determined that it makes sense and have started using it in production. Now is the time to document it.


shlomi-noach:
Call for review

GuptaManan100 (approved):

PR #1209, opened 2022-10-31
Use https protocol in git clone for operator quick start

spron-in:
By default github SSH is not configured and it failed for me. But HTTPS always works. Can HTTPS be used by default?

Signed-off-by: Sergey Pronin <spron-in@users.noreply.github.com>

### DIFF
--- a/content/en/docs/12.0/get-started/operator.md
+++ b/content/en/docs/12.0/get-started/operator.md
@@ -29,7 +29,7 @@ Before we get started, let’s get a few pre-requisites out of the way:
 Change to the operator example directory:
 
 ```bash
-git clone git@github.com:vitessio/vitess.git
+git clone https://github.com/vitessio/vitess
 cd vitess/examples/operator
 ```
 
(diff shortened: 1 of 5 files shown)


deepthi:
@spron-in any way of cloning a git repo works **if your environment is configured correctly**. This has nothing to do with vitess, it's a question of local setup.
So, you can do `ssh`, `https`, `gh` (CLI client) or even `git://`.
Rather than changing from `ssh` to `https`, would you be willing to change this PR to state that any method of cloning a git repo will work? Or at least show both examples - ssh and https.
Otherwise, one of the maintainers can do it.

Reference: https://git-scm.com/docs/git-clone#_git_urls

mattlord (approved):
Thank you, @spron-in ! ❤️ 

I do agree that using https by default is generally better because it requires less git/github setup (especially in "getting started" / "kicking the tires" places like this). I can see that we already use https in a number of places and git in others on the website so I will defer moving all `git clone` commands to https everywhere.

PR #1215, opened 2022-11-07
Archive `v12`

frouioui:
This Pull Request moves the `v12.0` documentation to the archive directory. `v12.0` has reached EOL during the last week of October 2022 and is thus no longer supported.


nate-double-u (approved):
Thanks for this @frouioui, and thanks for going through and fixing up those couple of file's links.

This /lgtm to me regarding the v12 archive. I'll leave it to someone else to merge as I'm not sure of the desired timing.

PR #1220, opened 2022-11-14
Change instructions for local install

rsajwani:
Signed-off-by: Rameez Sajwani <rameezwazirali@hotmail.com>

There is a a bug in v15.0 where local install fails when it tries to local vtadmin web locally. [PR#11683](https://github.com/vitessio/vitess/pull/11683) will fix the issue but we need to reflect these changes in our local install instructions as well. This PR focus on changing the instructions as needed. 

[Issue#11679](https://github.com/vitessio/vitess/issues/11679)

### DIFF
--- a/content/en/docs/15.0/get-started/local.md
+++ b/content/en/docs/15.0/get-started/local.md
@@ -85,11 +85,11 @@ Download the [latest binary release](https://github.com/vitessio/vitess/releases
 **Notes:**
 
 * Release 15.0 has a bug because of which the local example fails when try to run vtadmin web. [Issue#11679](https://github.com/vitessio/vitess/issues/11679)
-* Please use release 15.0.1 instead.
+* Please use release [15.0.1](https://github.com/vitessio/vitess/releases/tag/v15.0.1) instead.
 
 ```sh
-version=15.0.0
-file=vitess-${version}-e8c7e27.tar.gz
+version=15.0.1
+file=vitess-${version}-13ee9c8.tar.gz
 wget https://github.com/vitessio/vitess/releases/download/v${version}/${file}
 tar -xzf ${file}
 cd ${file/.tar.gz/}
--- a/content/en/docs/15.0/get-started/local.md
+++ b/content/en/docs/15.0/get-started/local.md
@@ -80,11 +80,16 @@ sudo setenforce 0
 
 ## Install Vitess
 
-Download the [latest binary release](https://github.com/vitessio/vitess/releases) for Vitess on Linux. For example with Vitess 13:
+Download the [latest binary release](https://github.com/vitessio/vitess/releases) for Vitess on Linux. For example with Vitess 15:
+
+**Notes:**
+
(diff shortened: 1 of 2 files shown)


content/en/docs/15.0/get-started/local.md
```diff
@@ -80,11 +80,16 @@ sudo setenforce 0
...
+**Notes:**
+
+* Release 15.0 has a bug because of which the local example fails when try to run vtadmin web. [Issue#11679](https://github.com/vitessio/vitess/issues/11679)
+* Please use release 15.0.1 instead.
 
 ```sh
-version=13.0.0
-file=vitess-${version}-bc4a960.tar.gz
+version=15.0.0
+file=vitess-${version}-e8c7e27.tar.gz
```
frouioui: I think it should be `15.0.1` instead of `15.0.0`

frouioui: After we have released `v15.0.1` we can edit this PR to change `file=vitess-${version}-e8c7e27.tar.gz` to use the proper SHA. However, this should be changed for future releases, we should find a way to automatise this.

content/en/docs/16.0/get-started/local.md
```diff
@@ -80,11 +80,11 @@ sudo setenforce 0
...
 ## Install Vitess
 
-Download the [latest binary release](https://github.com/vitessio/vitess/releases) for Vitess on Linux. For example with Vitess 13:
+Download the [latest binary release](https://github.com/vitessio/vitess/releases) for Vitess on Linux. For example with Vitess 15:
 
 ```sh
-version=13.0.0
-file=vitess-${version}-bc4a960.tar.gz
+version=15.0.0
+file=vitess-${version}-e8c7e27.tar.gz
```
frouioui: Same comment as in the `15.0` docs.

content/en/docs/16.0/get-started/local.md
```diff
@@ -80,11 +80,11 @@ sudo setenforce 0
 
 ## Install Vitess
 
-Download the [latest binary release](https://github.com/vitessio/vitess/releases) for Vitess on Linux. For example with Vitess 13:
+Download the [latest binary release](https://github.com/vitessio/vitess/releases) for Vitess on Linux. For example with Vitess 15:
```
frouioui, line 83: We should use the same "Notes" that you added in the `15.0` docs here, those:

```md
**Notes:**

* Release 15.0 has a bug because of which the local example fails when try to run vtadmin web. [Issue#11679](https://github.com/vitessio/vitess/issues/11679)
* Please use release 15.0.1 instead.
```

frouioui (commented):
LGTM, left a few comments that need to be addressed. Not approving this as we have to wait until we do the v15.0.1 release to merge it.

content/en/docs/15.0/get-started/local.md
GuptaManan100: Yep, I agree, it shouldbe v15.0.1 since we say above don't use 15.0.0

GuptaManan100 (dismissed):

frouioui (approved):

PR #1271, opened 2022-12-05
document global keyspace flag

maxenglander:
Accompanies vitessio/vitess#11876.

# Description
    
Vitess supports globally routing queries to unique tables when the connection does not have a default keyspace and the tables in the `FROM` list are not keyspace-qualified.
    
This adds support for explicitly requesting global routing:
    
* Adds a VTGate flag --global-keyspace that can be specified multiple times.
* When a connection uses a --global-keyspace, it behaves the same was as if no default keyspace is set on the connection.
* When a query for a table is qualified with a --global-keyspace, it is routed to a globally unique table, even if it is in a different keyspace than the current database.

# Use cases
    
This feature is useful for frameworks like Hibernate which are required by to set a database name on the connection. Those frameworks are unable to take advantage of global routing.

### DIFF
--- a/content/en/docs/16.0/reference/features/global-routing.md
+++ b/content/en/docs/16.0/reference/features/global-routing.md
@@ -4,12 +4,32 @@ weight: 23
 ---
 
 # Global Table Routing
-Vitess has an implicit feature of routing the queries to the appropriate keyspace based on the table specified in the `from` list.
-This differs from the standard mysql, in mysql unqualified tables will fail if the correct database is not set on the connection.
 
-This feature works only for unique table names provided in the [VSchema](https://vitess.io/docs/concepts/vschema/), and only when no default keyspace is set on the connection. One exception to the uniqueness rule is [Reference Tables](../../../user-guides/vschema-guide/advanced-vschema/#reference-tables) that share a name with an explicitly specified `source` table.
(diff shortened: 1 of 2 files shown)


content/en/docs/16.0/reference/features/global-routing.md
```diff
@@ -94,7 +118,12 @@ mysql> show columns from customer;
 ERROR 1105 (HY000): ambiguous table reference: customer
 ```
 
-With the default keyspace set to `customer` we can only query tables in `commerce` i.e `customer` and `corder`.
+<br/>
+
+#### With a default keyspace
+
+With the default keyspace set to `customer` we can only query keyspace-unqualified tables in `commerce` i.e `customer` and `corder`.
```
rsajwani, line 125: shouldn't it be 
"keyspace-unqualified tables in `customer`"
instead of 
"keyspace-unqualified tables in `commerce`"

rsajwani (commented):
Rest looks good to me.

maxenglander:
Returning this to draft status until there's more clarity around next steps for https://github.com/vitessio/vitess/pull/11938

deepthi:
Closing this because the code PR was closed without merging.

PR #1177, opened 2022-10-13
update vtctldclient docs

ajm188:
also went back and checked out `release-15` branch to make sure those docs were correct

PR #1318, opened 2022-12-22
Add tooling and makefile target for autogenerating `vtctldclient` for N versions

ajm188:
I ran this against #1177 and it produced no diffs. I then made some edits:

```
$ git diff content/
diff --git a/content/en/docs/15.0/reference/programs/vtctldclient/vtctldclient_AddCellsAlias.md b/content/en/docs/15.0/reference/programs/vtctldclient/vtctldclient_AddCellsAlias.md
index c900a8ac..d8ca2907 100644
--- a/content/en/docs/15.0/reference/programs/vtctldclient/vtctldclient_AddCellsAlias.md
+++ b/content/en/docs/15.0/reference/programs/vtctldclient/vtctldclient_AddCellsAlias.md
@@ -6,6 +6,8 @@ series: vtctldclient
 
 Defines a group of cells that can be referenced by a single name (the alias).
 
+This is some extra text in 15.0 docs.
+
 ### Synopsis
 
 Defines a group of cells that can be referenced by a single name (the alias).
diff --git a/content/en/docs/16.0/reference/programs/vtctldclient/vtctldclient_AddCellInfo.md b/content/en/docs/16.0/reference/programs/vtctldclient/vtctldclient_AddCellInfo.md
index c616a5f8..89df1a68 100644
--- a/content/en/docs/16.0/reference/programs/vtctldclient/vtctldclient_AddCellInfo.md
+++ b/content/en/docs/16.0/reference/programs/vtctldclient/vtctldclient_AddCellInfo.md
@@ -6,6 +6,8 @@ series: vtctldclient
 
 Registers a local topology service in a new cell by creating the CellInfo.
 
+This is some extra text in 16.0 docs.
+
 ### Synopsis
 
 Registers a local topology service in a new cell by creating the CellInfo
```

Ran the tool:

```
$ VITESS_DIR=$(realpath ../vitess) make vtctldclient-docs
```

And the diff was gone:

```
$ git diff content/ | wc -l
       0
```

### DIFF
--- a/tools/vtctldclientdocs/fsutil.go
+++ b/tools/vtctldclientdocs/fsutil.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2023 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package main
 
 import (
--- a/tools/vtctldclientdocs/fsutil.go
+++ b/tools/vtctldclientdocs/fsutil.go
@@ -1,0 +1,21 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"os"
+)
+
+func isDir(path string) error {
+	dir, err := os.Stat(path)
+	switch {
+	case errors.Is(err, os.ErrNotExist):
+		return err
+	case err != nil:
+		return fmt.Errorf("failed to stat %s: %w", path, err)
+	case !dir.IsDir():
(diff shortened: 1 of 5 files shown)


tools/vtctldclientdocs/fsutil.go
```diff
@@ -0,0 +1,21 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"os"
+)
+
+func isDir(path string) error {
```
rsajwani, line 25: minor: 
isDir by name should return true or false instead or err.

tools/vtctldclientdocs/main.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
```
rsajwani, line 27: I think i would be good if we have header/few lines which explain what this tool does.

tools/vtctldclientdocs/fsutil.go
ajm188: disagree; the error we wrap provides some extra info to the end user for _why_ things failed. if it's the function name leading to this, let's come up with a better name that implies an `error` return value

rsajwani (approved):

PR #1352, opened 2023-01-18
Docs: documenting --in-order-completion DDL strategy flag

shlomi-noach:
Documenting the new `--in-order-completion` DDL strategy flag introduced in https://github.com/vitessio/vitess/pull/12113

This PR affects these docs:

- DDL strategies
- Advanced usage


shlomi-noach:
relevant pages preview:

- https://deploy-preview-1352--vitess.netlify.app/docs/16.0/user-guides/schema-changes/ddl-strategy-flags/
- https://deploy-preview-1352--vitess.netlify.app/docs/16.0/user-guides/schema-changes/advanced-usage/

content/en/docs/16.0/user-guides/schema-changes/advanced-usage.md
```diff
@@ -153,3 +153,28 @@ $ vtctlclient -- ApplySchema --skip_preflight --sql "alter vitess_migration '3cc
...
+
+## Inter-dependent migrations
+
+It is possible to submit inter-dependent migrations within the same `ApplySchema` command, and have them complete in the correct order, even if they run concurrently. Examples for inter-dependent migrations:
+
+- Creating two new views, one of which reads from the other.
+- Adding a column to a table, and creating a new view that reads from that column.
+- Adding a column to a table, altering an existing view that reads from that table, to now read the new column.
+
+In the above examples there has to be a strict ordering to the migrations. You cannot just create a view that reads from a yet non-existent column.
```
mattlord, line 165: Technically you can, as far as MySQL is concerned, so it's incredibly nit-picky but "should not" is probably better than "can not".

content/en/docs/16.0/user-guides/schema-changes/advanced-usage.md
```diff
@@ -153,3 +153,28 @@ $ vtctlclient -- ApplySchema --skip_preflight --sql "alter vitess_migration '3cc
...
+
+`vitess` offers the `--in-order-completion` DDL strategy flag. It is the responsibility of the user to supply the migrations in a valid ordering, and it is `vitess`'s responsibility to _complete_ the migrations in that same order.
+
+Note that there can be scenarios with impossible ordering. Those hardly make sense in production, in the first place, and it is the user's responsibility to supply a sequence that works. When in doubt, it's advisable to submit migrations in stages: only apply one migration to completion, and then apply another.
+
+An example for in-order submission:
+
+```sh
+$ vtctlclient -- ApplySchema --skip_preflight --ddl_strategy='vitess --allow-concurrent --in-order-completion' --sql "create table t1 (id int primary key); create view v1 as select id from t1;" commerce
```
mattlord: We should use `vtctldclient` moving forward whenever we can (in some cases such as vreplication today, we cannot as they haven't been migrated).

content/en/docs/16.0/user-guides/schema-changes/advanced-usage.md
```diff
@@ -153,3 +153,28 @@ $ vtctlclient -- ApplySchema --skip_preflight --sql "alter vitess_migration '3cc
...
+```
+
+Notes:
+
+- `--allow-concurrent` is optional, but is likely to be the main use case for using in-order completion.
+- in-order completion also works with `--postpone-launch` and `--postpone-completion`.
```
mattlord: IMO we should use a note/info block:
```
{{< info >}}
- `--allow-concurrent` is optional, but is likely to be the main use case for using in-order completion.
 - in-order completion also works with `--postpone-launch` and `--postpone-completion`.
{{< /info >}}
```

content/en/docs/16.0/user-guides/schema-changes/ddl-strategy-flags.md
```diff
@@ -17,6 +17,11 @@ Vitess respects the following flags. They can be combined unless specifically in
 
 - `--fast-range-rotation`: when the migration runs on a table partitioned by `RANGE`, and the migration either runs a single `DROP PARTITION` or a single `ADD PARTITION`, and nothing other than that, then this flags instructs Vitess to run the `ALTER TABLE` statement directly against MySQL, as opposed to running an Online DDL with a shadow table. For `DROP PARTITION`, this flag is actually always desired, and will possibly become default/redundant in the future. If all conditions are indeed met, then the migration is not revertible.
 
+- `--in-order-completion`: a migration that runs with this DDL strategy flag may only complete if no prior migrations are still pending (pending means either `queued`, `ready` or `running` states). `--in-order-completion` considers the order by which migrations were submitted. Note that `--in-order-completion` still allows concurrency. In fact, it is designed to work with concurrent migrations. The idea is that as many migrations may run concurrently, but they way they finally _complete_ is in-order.
```
mattlord: `but they way they finally _complete_ is in-order` feels off. 

Recommendation:
```
The idea is that while many migrations may run concurrently, they must _complete_ in-order.
```
```

File: content/en/docs/16.0/user-guides/schema-changes/ddl-strategy-flags.md [hunk 5]
```diff
@@ -17,6 +17,11 @@ Vitess respects the following flags. They can be combined unless specifically in
...
+  - This lets the user submit multiple migrations which may have some dependencies (for example, introduce two views, one of which reads from the other). As long as the migrations are submitted in a valid order, the user can then expect `vitess` to complete the migrations successfully (and in that order).
+  - This strategy flag applies to any `CREATE|DROP TABLE|VIEW` statements, and to `ALTER TABLE` with `vitess|online` strategy.
+  - It _does not_ apply to `ALTER TABLE` in `gh-ost`, `pt-osc`, `mysql` and `direct` strategies.
```
mattlord: Tiny changes:
```
  - This strategy flag applies to any `CREATE|DROP TABLE|VIEW` statements, and to `ALTER TABLE` with the `vitess|online` strategy.
  - It _does not_ apply to `ALTER TABLE` when using the `gh-ost`, `pt-osc`, `mysql`, or `direct` strategies.
```

mattlord (approved):
LGTM! Only had minor nits.

content/en/docs/16.0/user-guides/schema-changes/advanced-usage.md
shlomi-noach: @mattlord you can't strictly `create view v as select * from non_existent_table` ; you _can_ create `non_existent_table`, create the view, then drop the table; but you can't perform the original, can you? And that's actually what the restriction is about.

shlomi-noach: I'd like to rephrase and explain that better. A `create view v as select * from non_existent_table` is impossible to run. That is the type of query for which you want `--in-order` in the first place. If you have two queries:

```sql
create table t(id int primary key);
create view v as select id from t;
```

Then you cannot create the latter without the former. The point of `--in-order` is that `vitess` will complete the migrations in the same order they were given; so in the above, `vitess` completes `t` before `v`. The other way around would not work. So I think the strict statement is correct in this case.

content/en/docs/16.0/user-guides/schema-changes/advanced-usage.md
shlomi-noach: fixed

content/en/docs/16.0/user-guides/schema-changes/ddl-strategy-flags.md
shlomi-noach: fixed

content/en/docs/16.0/user-guides/schema-changes/advanced-usage.md
shlomi-noach: Now using `vtctldclient`

content/en/docs/16.0/user-guides/schema-changes/ddl-strategy-flags.md
shlomi-noach: fixed. Please feel free to use GitHub's "suggest" feature so I can commit directly :pray:

shlomi-noach:
https://github.com/vitessio/vitess/pull/12113 is merged (in v16) hence merging this doc page.

PR #1370, opened 2023-01-30
Hotfix: enable site wide search

### LINKED ISSUE #1355: [Feedback] Search Functionality
## Vitess.io Search
I know during one of the Vitess meetings the community has been asked for feedback on the search functionality. I'm having difficulty finding a vttablet option I know is in the documentation: `init_db_name_override`.  

Screenshot from 2023-01-18 11-56-46

## Google Search
When doing a google search for the phrase I get provided two pages, both of which could help provide the information I'm looking for. 

Screenshot from 2023-01-19 09-56-36

thisisobate:
### Issue
Lunr depends on whatever we give it as the index to function. In the case of Vitess, we only crawl the summary of each of the docs pages. The summary is usually the first 5 or more sentences of a page hence the rest of the page/doc is not in the index.

### Solution
To solve this issue, I  created a new field containing all the page contents. This enables a complete site-wide search for content in Vitess.io

### Future Improvements
To make the vitess.io search better, we need to improve two things:

1. Reduce the search results summary to 1 or 2 sentences
2. Highlight the search keyword in the search result

### Action Item

- [x] Create an issue on Vitess.io for these different search improvements.

Fixes #1355

### DIFF
--- a/layouts/_default/list.lunr.json
+++ b/layouts/_default/list.lunr.json
@@ -1,5 +1,5 @@
 {{- $index := slice -}}
 {{- range $page := $.Site.RegularPages -}}
-    {{- $index = $index | append (dict "title" $page.Title "version" (index (first 3 (split (delimit (split $page.RelPermalink "/") "," "") ",")) 2) "href" $page.Permalink  "summary" ($page.Summary | plainify | safeHTML) "path" ($page.Permalink | relURL)  ) -}}
+    {{- $index = $index | append (dict "title" $page.Title  "content" ($page.Content | plainify | safeHTML) "version" (index (first 3 (split (delimit (split $page.RelPermalink "/") "," "") ",")) 2) "href" $page.Permalink  "summary" ($page.Summary | plainify | safeHTML) "path" ($page.Permalink | relURL)  ) -}}
 {{- end -}}
 {{- $index | jsonify -}}
--- a/layouts/partials/javascript.html
+++ b/layouts/partials/javascript.html
@@ -83,6 +83,12 @@ $versions.current}}
             wildcard: lunr.Query.wildcard.TRAILING,
             presence: lunr.Query.presence.REQUIRED,
           });
+          this.field("content", {
+            boost: 20,
+            usePipeline: true,
+            wildcard: lunr.Query.wildcard.TRAILING,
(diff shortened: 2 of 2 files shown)


GuptaManan100 (approved):

RobertKielty:
@thisisobate Thanks for the quick turnaround on this!

On the preview deploy,  olap pages are now being found, as are pages with init_db_name_override.   

@FancyFane PTAL at https://deploy-preview-1370--vitess.netlify.app/ 
@orware would love to get your feedback on the workload search  

@thisisobate I have a UX nit. Now that the index holds _all the content_ the search functionality has more work to do and there is lag between starting the search and providing the result. 

Let's have the search event handler **immediately display a spinner animation** so that as soon as the search starts, the end-user gets immediate feedback that a search is taking place. What do you think? Doable?

nate-double-u:
> @thisisobate I have a UX nit. Now that the index holds all the content the search functionality has more work to do and there is lag between starting the search and providing the result.

> Let's have the search event handler immediately display a spinner animation so that as soon as the search starts, the end-user gets immediate feedback that a search is taking place. What do you think? Doable?

Is this something we can do in a follow up PR? If the UX update is more than a day's effort, maybe we get the functionality in first and then work to smooth out the UX experience.

orware:
@RobertKielty The search functionality seemed considerably better at including some of those pages that were previously not being included now in that preview URL so from my end I am seeing an improvement compared to before!

thisisobate:
> Is this something we can do in a follow up PR?

I agree with you @nate-double-u 

Let's tackle this in a follow-up PR. 
Thanks for the suggestion @RobertKielty

FancyFane:
Also providing feedback, the search functionality is now much better. Doing the search for `init_db_name_override` yielded ideal search results. Thank you for the quick turn around on this!!

nate-double-u:
Let's merge this in as is then, and we can open up a follow up issue.

deepthi:
Thank you @RobertKielty @thisisobate for fixing this! Looking forward to the UX improvement PR 😛

RobertKielty:
The following issues were created arising out of doing this work

#1371 
#1372
#1376

PR #1411, opened 2023-03-16
Update the local install guide for `v16.0.1` and `v17`

frouioui:
This PR is preparing the upcoming `v16.0.1` release and updates the local install guide which were outdated on both v16 and v17.


GuptaManan100 (approved):

frouioui:
Marking this as a draft as I would like to automate this to avoid this recurring issue in the future.

PR #1416, opened 2023-03-21
content/en/docs/18.0: clarify meaning of scatter queries

maxenglander:
As far as I can tell, the docs do not offer a clear definition of what constitutes a scatter query, and some parts of the docs are inaccurate in regard to this term.

This PR offers a definition of scatter queries, and cleans up one place that isn't accurate.


frouioui (commented):
looks good to me, however we need to copy this to the v18.0 docs.

harshit-gangal:
@maxenglander can you port this to v18 docs?

harshit-gangal (approved):

PR #1422, opened 2023-03-27
added source for images for connection pooling blog

harshit-gangal:
This PR adds the source of the images used in the blog


deepthi (approved):

deepthi:
Oh, I noticed that the attribution is missing on the last images (graphs). Can you add that one, and then this will be good to merge?

harshit-gangal:
> Oh, I noticed that the attribution is missing on the last images (graphs). Can you add that one, and then this will be good to merge?

I removed it on purpose. rather I linked the PlanetScale name to the website.

harshit-gangal:
> > Oh, I noticed that the attribution is missing on the last images (graphs). Can you add that one, and then this will be good to merge?
> 
> I removed it on purpose. rather I linked the PlanetScale name to the website.

I will merge this for now. If you think we should add that as well, will do it in a separate PR.

deepthi:
It might actually have been better to either put one line at the top saying "Originally published at ..." or a line at the bottom saying "All images used with permission from `link_to_original_blog_post`"
Also looks like we might need to add `Disallow: /blog/2023-03-27-connection-pooling-in-vitess/` to `robots.txt` to avoid being penalized by search engines for duplicate content.

harshit-gangal:
> It might actually have been better to either put one line at the top saying "Originally published at ..." or a line at the bottom saying "All images used with permission from `link_to_original_blog_post`"
> Also looks like we might need to add `Disallow: /blog/2023-03-27-connection-pooling-in-vitess/` to `robots.txt` to avoid being penalized by search engines for duplicate content.

I will add the line at top. sounds good.

PR #1435, opened 2023-04-04
Proposed `schemadiff` blog post

shlomi-noach:
Titled "'schemadiff: in-memory schema diffing, normalization, validation and manipulation'"

Randomly scheduled for April 15 (which is on a weekend) -- we'll change that per blog release schedule.


shlomi-noach:
Blog preview: ~~https://deploy-preview-1435--vitess.netlify.app/blog/2023-04-15-schemadiff/~~ https://deploy-preview-1435--vitess.netlify.app/blog/2023-04-24-schemadiff/

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+This approach has two advantages:
+
+1. If you're able to create a table in MySQL, that means it's valid. By the time you introspect `INFORMATION_SCHEMA` on that table, you may safely assume that, for example, keys only cover existing columns.
+2. `INFORMATION_SCHEMA` formalizes the majority of information. The data type is well defined. A column precision is an integer value. The text collation is a well known value.
+
+However, there are disadvantages, as well:
+
+- Complex expressions are left, well, complex. A `CREATE VIEW ...` statement is really just a long SQL text. Which tables or views are referenced in our `VIEW` definition? Which columns?
+- Not everything is normalized and formalized. We know `int(10)` and `int(11)` are really the same. But as MySQL goes, the column type is different. Contract this with `varchar(10)` and `varchar(11)`, where the type is truly different. It still takes some higher level logic to understand what is a real diff and what isn't.
```
dbussink: Maybe good to call out here that this is deprecated for MySQL 8.0 and link to that it doesn't mean anything anymore? Feels like it's maybe something people don't really all know yet since I see ints with lengths still in so many places.

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+```
+
+The first thing to note in the above examples is that everything takes place purely within `go` space, and MySQL is not involved. `schemadiff` is purely programmatic, and makes heavy use of Vitess' [`sqlparser`](https://github.com/vitessio/vitess/tree/main/go/vt/sqlparser) library.
+
+
+## sqlparser
+
+Vitess is a sharding and infrastructure framework running on top of MySQL, and masquerades as a MySQL server to route queries into relevant shards. It thus obviously must be able to parse MySQL's SQL dialect. [`sqlparser`](https://github.com/vitessio/vitess/tree/main/go/vt/sqlparser) is the Vitess library that does so.
+
+`sqlparser` utilizes a [classic yacc file](https://github.com/vitessio/vitess/blob/main/go/vt/sqlparser/sql.y) to parse SQL into an Abstract Syntax Tree (AST), with `golang` structs generated and populated by the parser. For example, a SQL `CREATE TABLE` statement is parsed into a [`CreateTable`](https://github.com/vitessio/vitess/blob/157857a4c5ee6dcb44e7de08894db8334750746e/go/vt/sqlparser/ast.go#L509-L517) instance:
```
dbussink: Maybe useful to also link to https://en.wikipedia.org/wiki/Yacc here?

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+	id int primary key,
+	i int(11) default null,
+	v varchar(32) collate utf8mb4_0900_ai_ci
+) default charset=utf8mb4 collate=utf8mb4_0900_ai_ci
+```
+
+There's a few things to normalize here. Inputs can come in different shapes and sizes, and still mean the same thing. Even MySQL itself presents different output for textual columns based on which version the table was created in. In the above table definition, we can normalize the following:
+
+- The canonical way to declare a `primary key` is as an index definition, not as part of the column definition.
+- `int(12)` is just an `int`. `12` does not matter. Interestingly, some ORMs do have special treatment for `int(1)`, as an indication to a boolean value. `schemadiff` accommodates that.
```
dbussink: The example has `int(11)` but you talk about `int(12)` here.

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+- It looks like `i` is `NULL`able (because it does not say `not null`). Which means its default value is `null` even if we don't explicitly say that.
+- `v`'s collation, and thereby character set, agree with the table's collation. It can be removed.
```
dbussink: Should we indicate that this is a choice we made for `schemadiff` to have it be simpler for the user to read. Vs. MySQL, which is often very verbose here in `show create table`?

content/en/blog/2023-04-15-schemadiff.md
shlomi-noach: Added link to https://dev.mysql.com/worklog/task/?id=13127 ; it's the best I could find that talks not only about `ZEROFILL` but also about precision.

content/en/blog/2023-04-15-schemadiff.md
shlomi-noach: Done

content/en/blog/2023-04-15-schemadiff.md
shlomi-noach: Fixed

content/en/blog/2023-04-15-schemadiff.md
shlomi-noach: Done, in next paragraph.

shlomi-noach:
I will make minor edits to the code samples per https://github.com/vitessio/vitess/pull/12551 and https://github.com/vitessio/vitess/issues/12885

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
+---
+author: 'Shlomi Noach'
+date: 2023-04-15
+slug: '2023-04-15-schemadiff'
+tags: ['Vitess','MySQL', 'DDL', 'schema']
+title: 'schemadiff: in-memory schema diffing, normalization, validation and manipulation'
```
deepthi: Will be nice to capitalize some of the words. Suggestion follows.

deepthi:
```suggestion
title: 'schemadiff: Vitess In-memory Schema Diffing, Normalization, Validation and Manipulation'
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+description: 'Introducing schemadiff, a best kept secret library available in Vitess'
+---
+
+Introducing `schemadiff`, a best kept secret internal `Vitess` library. At its core, `schemadiff` is a declarative, programmatic library that can produce an SQL diff of entities: two tables, two views, or two full blown database schemas. But it then goes beyond that to normalize, validate, export, and even _apply_ schema changes, all declaratively and without having to use a MySQL server. Let's dive in to understand its functionality and capabilities.
```
deepthi: This phrase is somewhat awkward. The problem is that when you say `best kept secret`, secret is used as a noun, but when you say `secret internal library` it's an adjective, and mixing the two doesn't flow properly.
Some suggestions:
- Introducing `schemadiff`, one of the best kept secrets in `Vitess`
- Introducing `schemadiff`, a hitherto secret internal library in `Vitess`
- Introducing `schemadiff`, an internal library in `Vitess` that has been one of its best-kept secrets until now

deepthi: `an SQL diff` versus `a SQL diff`.
Depends on whether SQL is pronounced `es-queue-elle` in which case `an` is correct. If it is pronounced `sequel` then `a` is correct. 
You can avoid this whole thing by saying something like `a diff in SQL format`

deepthi:
```suggestion
Introducing `schemadiff`, a best kept secret internal `Vitess` library. At its core, `schemadiff` is a declarative, programmatic library that can produce an SQL diff of two entities: tables, views, or full blown database schemas. But it then goes beyond that to normalize, validate, export, and even _apply_ schema changes, all declaratively and without having to use a MySQL server. Let's dive in to understand its functionality and capabilities.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Some tech specs
+
+`schemadiff` supports MySQL `8.0` dialect and functionality. It is completely in-memory and does not use any MySQL server or MySQL code. It applies MySQL-compatible logic and rules to validate and apply schema changes. It uses a declarative approach but also works with imperative commands.
+
+## Diff preamble
```
deepthi: Do you mean `preamble` here in the sense of `Introduction`? My suggestion is to call it `Motivation` or `Objectives`.

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+`schemadiff`, as its name suggests, began as a diffing library. The objectives were:
+
+1. Given two table definitions, what schema changes (DDLs) would we need to apply on the 1st, so it looks like the 2nd?
+2. Given two schemas (aka databases), that include tables and views, what schema changes (DDLs) would we need to apply on the 1st, so it looks like the 2nd?
```
deepthi:
```suggestion
1. Given two table definitions, what schema changes (DDLs) would we need to apply on the first, so it looks like the second?
2. Given two schemas (aka databases), that include tables and views, what schema changes (DDLs) would we need to apply on the first, so it looks like the second?
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+As a use case for (1), consider [declarative Vitess migrations](https://vitess.io/docs/16.0/user-guides/schema-changes/declarative-migrations/). Consider that you might want to tell your database: "Here's a table, please make it look like so". It's the database's job to determine whether this table exists in the first place or not. If not, then it must be created. If it exists, and already looks exactly like you want it to, geat, that's a no-op. But what if the current table looks somewhat different? What changes need to be applied on the existing table so as to make it look like your desired table? Specifically, what `ALTER TABLE` statement(s) (there are some rare scenarios where we might need to invoke more than one) do we need to invoke?
```
deepthi:
```suggestion
As a use case for (1), consider [declarative Vitess migrations](https://vitess.io/docs/16.0/user-guides/schema-changes/declarative-migrations/). Consider that you might want to tell your database: "Here's a table, please make it look like so". It's the database's job to determine whether this table exists in the first place or not. If not, then it must be created. If it exists, and already looks exactly like you want it to, great, that's a no-op. But what if the current table looks somewhat different? What changes need to be applied on the existing table so as to make it look like your desired table? Specifically, what `ALTER TABLE` statement(s) (there are some rare scenarios where we might need to invoke more than one) do we need to invoke?
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+There are two approaches to analyzing the diff of tables or schemas. One approach is to use a running MySQL server. Deploy the two schemas on that server, then investigate the `INFORMATION_SCHEMA` metadata tables and construct a model for the schema. `INFORMATION_SCHEMA` offers a formal breakdown such as the columns found in each table, what data type a column has, what default does it have, if any; which indexes are unique, which columns do they cover; what foreign key constraints on a table point to which parent table, what columns are covered; etc.
+
+This approach has two advantages:
+
+1. If you're able to create a table in MySQL, that means it's valid. By the time you introspect `INFORMATION_SCHEMA` on a table, validity is given. You may safely assume, for example, that keys only cover existing columns.
```
deepthi:
```suggestion
1. If you're able to create a table in MySQL, that means it's valid. By the time you introspect `INFORMATION_SCHEMA` on a table, validity is a given. You may safely assume, for example, that keys only cover existing columns.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+2. `INFORMATION_SCHEMA` formalizes the majority of information. The data type is well defined. A column precision is an integer value. The text collation is a well known value.
+
+However, there are disadvantages, as well:
+
+- Complex expressions are left, well, complex. A `CREATE VIEW ...` statement is really just a long SQL text. Which tables or views are referenced in our `VIEW` definition? Which columns?
+- Not everything is normalized and formalized. We know `int(10)` and `int(11)` are really the same. But as MySQL goes, the column type is different. Contract this with `varchar(10)` and `varchar(11)`, where the type is truly different. It still takes some higher level logic to understand what is a real diff and what isn't.
```
deepthi:
```suggestion
- Not everything is normalized and formalized. We know `int(10)` and `int(11)` are really the same. But as MySQL goes, the column type is different. Contrast this with `varchar(10)` and `varchar(11)`, where the type is truly different. It still takes some higher level logic to understand what is a real diff and what isn't.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+- MySQL allows schema inconsistencies. It is possible, in MySQL, to have an "orphaned" `VIEW`, where the tables/views on which it relies, do not exist. Reading something from MySQL doesn't fully mean it's valid.
```
deepthi:
```suggestion
- MySQL allows schema inconsistencies. It is possible, in MySQL, to have an "orphaned" `VIEW`, where the tables/views on which it relies, do not exist. Reading something from MySQL doesn't mean it's really valid.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+- Last, and most impactful of all, is the operational overhead. To diff two schemas, you need to deploy two schemas, then read back the metadata for all the tables, views, indexes, constraints, etc. You need to deploy a MySQL server, likely in a non-production environment. You will need to start the server, deploy the changes, read back, shutdown the server. This is a heavyweight operation.
+  Even more, if you want to validate your _diff_, you probably want to deploy it on the running server, then re-read the result, compare it with what you were expecting to find. This is more so heavyweight.
```
deepthi:
```suggestion
  And if you then want to validate your _diff_, you probably want to deploy it on the running server, then re-read the result, compare it with what you were expecting to find. This is even more heavyweight.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+	And, if you wanted to resolve the _order_ by which the changes must take place (some scenarios dictate a specific order, a topic for a future post), then you'd need to solicit confirmation from MySQL by applying changes on the server, regressively.
```
deepthi:
```suggestion
	And, if you want to compute the _order_ in which the changes must take place (some scenarios dictate a specific order, a topic for a future post), then you'd need to solicit confirmation from MySQL by applying changes on the server, regressively.
```

deepthi: I don't understand the use of the word `regressively` here. What is it supposed to convey?

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+The second is the in-memory, programmatic approach. Instead of relying on `INFORMATION_SCHEMA`, we rely on the SQL of the schema, i.e. on the `CREATE TABLE` and `CREATE VIEW` statements themselves. This poses a few challenges of its own:
+
+- First and foremost, you must be able to parse and analyze all statements. This includes a `CREATE TABLE` that has a `CHECK CONSTRAINT` with a complex expression, or a sub-partitioning scheme using functions over columns. Or the ability to fully parse a `CREATE VIEW` statement. Or a complex `GENERATED` column expression, etc.
```
deepthi:
```suggestion
- First and foremost, you must be able to parse and analyze all statements. This includes a `CREATE TABLE` that has a `CHECK CONSTRAINT` with a complex expression, or a sub-partitioning scheme using functions over columns. You also need the ability to fully parse a `CREATE VIEW` statement, a complex `GENERATED` column expression, etc.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+- You must be able to accommodate different equivalent definitions. For example, `create table t (id int primary key)` is equivalent to `create table t(id int, primary key (id))`. Even MySQL's own `SHOW CREATE TABLE` output may present different results depending on when/where the table was created.
+- Not having a MySQL server to validate correctness of the initial schema and the generated diffs means the application must implement that logic.
```
deepthi:
```suggestion
- Not having a MySQL server to validate correctness of the initial schema and the generated diffs means that the application must implement that logic.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+`schemadiff` uses the programmatic approach. Let's see some sample code and continue to discuss its internals.
```
deepthi:
```suggestion
`schemadiff` uses the programmatic approach. Let's look at some sample code first and then we'll move on to discuss its internals.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+}
+```
+
+And this is the resulting diff:
+
+```sql
+ALTER TABLE `t` MODIFY COLUMN `id` bigint, ADD KEY `name_idx` (`name`(16))
+```
+
+There's different ways to generating the diffs. Here's a quick shortcut to achieving the same:
```
deepthi:
```suggestion
There are multiple ways of generating the diff. Here's a quick shortcut to achieving the same:
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+diffs, err := DiffSchemasSQL("create ...", "create ...", hints)
+...
+```
+
+The main thing to note in the above examples is that everything takes place purely within `go` space, and MySQL is not involved. `schemadiff` is purely programmatic, and makes heavy use of Vitess' [`sqlparser`](https://github.com/vitessio/vitess/tree/main/go/vt/sqlparser) library.
+
+
+## sqlparser
+
+Vitess is a sharding and infrastructure framework running on top of MySQL, and masquerades as a MySQL server to route queries into relevant shards. It thus obviously must be able to parse MySQL's SQL dialect. [`sqlparser`](https://github.com/vitessio/vitess/tree/main/go/vt/sqlparser) is the Vitess library that does so.
```
deepthi:
```suggestion
Vitess is a sharding and management framework running on top of MySQL, which masquerades as a MySQL server to route queries to relevant shards. It thus obviously must be able to parse MySQL's SQL dialect. [`sqlparser`](https://github.com/vitessio/vitess/tree/main/go/vt/sqlparser) is the Vitess library that does so.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+type TableSpec struct {
+	Columns         []*ColumnDefinition
+	Indexes         []*IndexDefinition
+	Constraints     []*ConstraintDefinition
+	Options         TableOptions
+	PartitionOption *PartitionOption
+}
+```
+
+You can already see how AST helps us in analyzing a table's definition. As a very simple illustration, imagine we have two tables we want to diff. Say we want to find whether the two have a different set of columns (let's ignore ordering for now). How would we do that?
```
deepthi:
```suggestion
You can already see how AST helps us in analyzing a table's definition. As a very simple illustration, imagine we have two tables we want to diff. Say we want to determine whether the two have a different set of columns (let's ignore ordering for now). How would we do that?
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+We can programmatically iterate over `range table1.TableSpec.Columns` in each of the tables. We can do a full drill down of all the details in a `ColumnDefinition`. Or, we can take a shortcut. `schemadiff` uses an optimistic approach: most of the schema is likely to be identical. It first attempts to compare components as a whole. If they components are identical as a whole, we can proceed to drill down.
```
deepthi:
```suggestion
We can programmatically iterate over `range table1.TableSpec.Columns` in each of the tables. We can do a full drill down of all the details in a `ColumnDefinition`. Or, we can take a shortcut. `schemadiff` uses an optimistic approach: most of the schema is likely to be identical. It first attempts to compare components as a whole. If the components are identical as a whole, we can proceed to drill down.
```

deepthi: If the components are NOT identical as a whole?
If they are identical, there is no diff.

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+How do you "compare a component as a whole"? `sqlparser` not only parses SQL, it also exports AST as SQL. We can, for example, run:
+
+```go
+  s1 := sqlparser.CanonicalString(table1.TableSpec.Columns[i])
+  s2 := sqlparser.CanonicalString(table2.TableSpec.Columns[i])
+```
+
+We can thus compare any two _nodes_. If they export into identical strings, then they are equal and there is no need to drill down. `sqlparser`'s AST also comes with an auto-generated set of deep-equals comparison and deep-copy functionality.
```
deepthi: you may take or leave this suggestion, it's debatable
`deep-equals` -> `DeepEquals`
`deep-copy` -> `DeepCopy`

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+Continuing our simple illustration, if `table2` has a column called `info`, which we cannot find in `table1`, then our table diff, our `ALTER TABLE` statement, should include an `ADD COLUMN <column definition>` statement. What's the content of the column definition? It is trivially the `CanonicalString()` of the extra column object.
```
deepthi:
```suggestion
Continuing our simple illustration, if `table2` has a column called `info`, which we cannot find in `table1`, then our table diff, i.e. our `ALTER TABLE` statement, should include an `ADD COLUMN <column definition>` statement. What's the content of the column definition? It is trivially the `CanonicalString()` of the extra column object.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+`sqlparser` is devoid of semantic context. It merely deals with a programmatic reflection of SQL. It does not know if a certain table exists, nor does it care about type conversions in a `SELECT` query. As long as the syntax is valid, it is satisfied.
```
deepthi: Why is it relevant to talk about type conversions in `SELECT` in this context? If it isn't relevant we can remove that phrase.

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+```sql
+CREATE TABLE `invalid` (
+	`id` bigint,
+	`title` varchar(64),
+	`title` tinytext,
+	PRIMARY KEY (`val`)
+);
+```
+
+The above table is _syntactically_ valid, but _semantically_ invalid. There are two columns that go by the same name (`title`), and an index that covers a non-existent column (`val`). These are but two out of many possible errors. A `SHOW CREATE TABLE` would never produce such an output, of course. But we can't assume the output comes from MySQL. If we don't validate the input, and attempt to produce a diff, results could be unexpected.
```
deepthi:
```suggestion
The above table is _syntactically_ valid, but _semantically_ invalid. There are two columns that go by the same name (`title`), and an index that covers a non-existent column (`val`). These are but two out of many possible errors. A `SHOW CREATE TABLE` would never produce such an output, of course. But we can't assume the output comes from MySQL. If we don't validate the input, and attempt to produce a diff, the results could be unexpected.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Validation
+
+The basic validation premise is that everyone should want their schema to be valid. There's no point in using invalid schemas, because, at the end of they day, you want to be able to deploy those schemas on real MySQL servers.
```
deepthi:
```suggestion
The basic premise of validation is that all users want their schemas to be valid. There's no point in using invalid schemas, because, at the end of they day, you want to be able to deploy those schemas on real MySQL servers.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+To that effect, `schemadiff` enforce validation upon loading a new schema. As we see later on, it also enforces validation upon applying changes. If you try to load an invalid schema as in the above, `schemadiff` returns an informative error. You can't have two columns of same name. Your keys may only cover existing columns. A collation name must be recognized. A `GENERATED` column must refer to existing columns. A `FOREIGN KEY` constraint must reference existing columns in existing tables, respective to the local columns count and data types, etc. Circular `FOREIGN KEY` dependencies are not allowed.
```
deepthi:
```suggestion
To that effect, `schemadiff` enforces validation upon loading a new schema. As we will see later on, it also enforces validation upon applying changes. If you try to load an invalid schema as in the above, `schemadiff` returns an informative error. You can't have two columns of same name. Your keys may only cover existing columns. A collation name must be recognized. A `GENERATED` column must refer to existing columns. A `FOREIGN KEY` constraint must reference existing columns in existing tables, respective to the local columns count and data types, etc. Circular `FOREIGN KEY` dependencies are not allowed.
```

deepthi: Can you clarify this phrase? 
```
respective to the local columns count and data types
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+Even more interesting is view validation. `schemadiff` offers a complete validation over view definitions: all referenced tables/views must exist. Circular dependency is not allowed. Referenced columns are validated to exist. Unqualified column names are ensured not to be ambiguous.
```
deepthi:
```suggestion
Even more interesting is validation of views. `schemadiff` offers a complete validation over view definitions: all referenced tables/views must exist. Circular dependency is not allowed. Referenced columns are validated to exist. Unqualified column names are ensured to be unambiguous.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Diff hints
+
+There are nuances. What do you do about different `AUTO_INCREMENT` values? These are not strictly part of the schema, but may or may not cause impact to production if you apply or not apply them. How do you deal with renamed columns? What about constraint names, which, per ANSI SQL, must be unique across the schema? Do you mind if key definitions are in different order?
```
deepthi:
```suggestion
There are nuances. What do you do about different `AUTO_INCREMENT` values? These are not strictly part of the schema, but may or may not cause impact to production if you apply or not apply them. How do you deal with renamed columns? What about constraint names, which, per ANSI SQL, must be unique across the schema? Do you care if key definitions are in a different order?
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+```sql
+create table what_can_be_normalized (
+	id int primary key,
+	i int(12) default null,
+	v varchar(32) collate utf8mb4_0900_ai_ci
+) default charset=utf8mb4 collate=utf8mb4_0900_ai_ci
+```
+
+There's a few things to normalize here. Inputs can come in different shapes and sizes, and still mean the same thing. Even MySQL itself presents different output for textual columns based on which version the table was created in. In the above table definition, we can normalize the following:
```
deepthi:
```suggestion
There are a few things to normalize here. Inputs can come in different shapes and sizes, and still mean the same thing. Even MySQL itself presents different output for textual columns based on which version the table was created in. In the above table definition, we can normalize the following:
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+CREATE TABLE `what_can_be_normalized` (
+	`id` int,
+	`i` int,
+	`v` varchar(32),
+	PRIMARY KEY (`id`)
+) CHARSET utf8mb4,
+  COLLATE utf8mb4_0900_ai_ci;
+```
+
+`schemadiff` chooses to normalize into the shortest form possible, often shorter than MySQL's own `SHOW CREATE TABLE` output. Normalization ensures we converge onto a single, canonical, representation of a table, view, or schema.
```
deepthi:
```suggestion
`schemadiff` chooses to normalize into the shortest form possible, often shorter than MySQL's own `SHOW CREATE TABLE` output. Normalization ensures we converge onto a single, canonical representation of a table, view, or schema.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Beyond diff
+
+Up till now, we saw that we can load a schema into `schemadiff`. `schemadiff` would normalize the schema, validate it, and throw an error if it's invalid. We can load a 2nd schema, and if all goes well we can diff the two, and grab the _diffs_.
```
deepthi:
```suggestion
Up until now, we saw that we can load a schema into `schemadiff`. `schemadiff` will normalize the schema, validate it, and throw an error if it's invalid. We can load a second schema, and if all goes well we can diff the two, and grab the _diffs_.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+What then? Do we trust those diffs? Will they result in the correct schema? Will they result in a valid schema?
+
+It looks like to validate those diffs we'd need to run a MySQL server, deploy the 1st schema, apply those diffs (`CREATE`, `ALTER`, `DROP` statements), then re-read the schema and compare it with the one we aimed for. But, this brings back the dependency on a MySQL server, something we wanted to avoid in the first place.
```
deepthi:
```suggestion
It seems that in order to validate those diffs we'd need to run a MySQL server, deploy the first schema, apply those diffs (`CREATE`, `ALTER`, `DROP` statements), then re-read the schema and compare it with the one we aimed for. But, this brings back the dependency on a MySQL server, something we wanted to avoid in the first place.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+applied, err := schema1.Apply(diffs)
+require.NoError(t, err)
+
+appliedDiff, err := schema2.Diff(applied, hints)
+require.NoError(t, err)
+assert.Empty(t, appliedDiff)
+assert.Equal(t, schema2.ToQueries(), applied.ToQueries())
+```
+
+`schemadiff` lets us take those _diffs_ and `Apply()` them onto a schema (`schema1`), resulting with a new schema. We expect that result to be equal in content to the expected (`schema2`) schema. We also expect `schemadiff` itself to agree the two are identical.
```
deepthi:
```suggestion
`schemadiff` lets us take those _diffs_ and `Apply()` them onto a schema (`schema1`), resulting in a new schema. We expect that result to be equal in content to the expected (`schema2`) schema. We also expect `schemadiff` itself to agree that the two are identical.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+`Apply()` makes an in-memory schema change. You may `CREATE TABLE` or `ALTER TABLE` or `DROP VIEW`, all programmatically, and all using AST structures such as `CreateTable`, `AlterTable`, or `DropView`, etc.
+
+`Apply()` validates the requested changes on the fly, as well as on the resulting schema. It adheres to MySQL rules. For example:
+
+- If the diff has a `AddColumn` AST struct, verify upfront that no column exists by same name.
```
deepthi:
```suggestion
- If the diff has an `AddColumn` AST struct, verify upfront that no column exists with the same name.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+- If there's an `AddKey`, validate that the columns specified by the key, do in fact exist.
```
deepthi:
```suggestion
- If there's an `AddKey`, validate that the columns specified by the key do in fact exist.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+- If there's an `AddKey` and no index name specified, generate one, compatible with MySQL naming.
+- Conversely, if there's a `DropColumn`, and some keys exist that actually cover that column, remove the column from those keys; any key that is left without any covered columns is dropped (this is the standard MySQL behavior).
+- You shouldn't be able to `DropColumn` if there's a `FOREIGN KEY` referencing that column, though.
+- The list goes on.
+
+As we can see, there's a lot of validations involved. Some of them actually depend on each other! Say we drop a column as well as a `FOREIGN KEY` that references that column. MySQL-wise this is a single `ALTER TABLE` operation. But, programmatically, `schemadiff` needs to first remove the `FOREIGN KEY`, and then remove the column. The reverse order is invalid. `schemadiff` resolves the correct order of operations inside a table.
```
deepthi:
```suggestion
As we can see, there are a lot of validations involved. Some of them actually depend on each other! Say we drop a column as well as a `FOREIGN KEY` that references that column. MySQL-wise this is a single `ALTER TABLE` operation. But, programmatically, `schemadiff` needs to first remove the `FOREIGN KEY`, and then remove the column. The reverse order is invalid. `schemadiff` computes the correct order of operations for a table.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+But then, it also resolves the correct order of operations in the grand context of the schema. Consider this diff:
```
deepthi: I've suggested replacing `resolves` with `computes` everywhere. You may choose to keep `resolves`. The most important thing is to use the same term everywhere consistently.

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+```sql
+create view v1 as select * from v2;
+create view v2 as select * from t;
+```
+
+Alphabetically, `v1` comes before `v2`. But, to apply these two diffs, we must first create `v2`, then `v1`. The reverse order is invalid. This time it's invalid not only programmatically in `schemadiff`, but also when applied to MySQL itself.
+
+`schemadiff` maintains the hierarchical ordering of all tables and views, based on `FOREIGN KEY` and view definitions. There's a specific order for `CREATE` statements, and a specific (reverse) order for `DROP` statements. There may actually be conflicts between the diffs, but that's the topic of a different blog post.
```
deepthi:
```suggestion
`schemadiff` maintains the hierarchical ordering of all tables and views, based on `FOREIGN KEY` and view definitions. There's a specific order for `CREATE` statements, and a specific (reverse) order for `DROP` statements. There may actually be conflicts between the diffs, but that's a topic for a different blog post.
```

content/en/blog/2023-04-15-schemadiff.md
```diff
@@ -0,0 +1,295 @@
...
+
+Because this runs in GitHub CI, we take advantage of a running MySQL server. We perform the test in-memory, and then also on MySQL itself, and so we have an authoritative validation for the `Apply()` functionality. For fun and glory, we also generate the reverse diff. This test cross validates the programmatic approach, the MySQL schema, and actual MySQL schema changes, and expects full compliance between all.
+
+## Even beyond
+
+The declarative approach is challenging. It requires reverse-engineering of a hidden sequence of changes. Sometimes these changes have inter-dependency between them. We will discuss this further in a future post.
```
deepthi:
```suggestion
The declarative approach is challenging. It requires reverse-engineering of a hidden sequence of changes. Sometimes these changes are inter-dependent. We will discuss this further in a future post.
```

deepthi (commented):
I like it! Very well-structured and informative.
Lots of wording / re-phrasing / grammar suggestions though.

deepthi:
@shlomi-noach can you rename the file and change the date so that we can plan to publish this on April 24, Monday? I assume all reviews can be addressed by then.

content/en/blog/2023-04-15-schemadiff.md
shlomi-noach: Replaced with 

> Introducing schemadiff, an internal library in Vitess that has been one of its best-kept secrets until now

shlomi-noach: fixed

content/en/blog/2023-04-15-schemadiff.md
shlomi-noach: I did mean `preamble` in the sense of `introduction`. Replaced with `objectives`

content/en/blog/2023-04-15-schemadiff.md
shlomi-noach: Wrong word. Replaced with `iteratively`. The idea that you'd have to try an ordering combination of migrations, back off if any failed, and attempt another order combination. Does that make more sense now?

content/en/blog/2023-04-15-schemadiff.md
shlomi-noach: Good catch. Fixed.

content/en/blog/2023-04-15-schemadiff.md
shlomi-noach: I think we should keep the current phrasing. That's because `DeepEquals` suggests there's actually a function called by that name; but that's not the case. There are multiple auto-generated `.*Equals.*` functions, named per entity. Likewise for Copy.

content/en/blog/2023-04-15-schemadiff.md
shlomi-noach: You're right. Replaced with a better example:

> It does not know if a certain table exists, or if a foreign key references valid columns.

content/en/blog/2023-04-15-schemadiff.md
shlomi-noach: Replaced with:

> and foreign key columns mast have matching data types

This is about a MySQL constraint where if you create a foreign key on a column, then in both parent and child tables, the column must have the exact same data type. If you have a foreign key over two columns, then each of the columns must respectively have the same data type in the parent and child tables.

content/en/blog/2023-04-15-schemadiff.md
shlomi-noach: You left one out :wink: , replaced

shlomi-noach:
@deepthi thank you, I believe I've addressed all of your comments. File is renamed and edited to be published on April 24th.

The new preview link is now: https://deploy-preview-1435--vitess.netlify.app/blog/2023-04-24-schemadiff/

shlomi-noach:
Updated code samples to reflect changes merged in https://github.com/vitessio/vitess/pull/12885

deepthi:
> @deepthi thank you, I believe I've addressed all of your comments. File is renamed and edited to be published on April 24th.
> 
> The new preview link is now: https://deploy-preview-1435--vitess.netlify.app/blog/2023-04-24-schemadiff/

The old file still needs to be deleted, there are now 2 files in this PR.
Rest LGTM, so I'm going to approve. You can plan to publish it next Monday.

deepthi (approved):

shlomi-noach:
> there are now 2 files in this PR.

That's bizzarre. I used `git mv`. Not sure how that file was left over. anyway, deleted.

shlomi-noach:
> You can plan to publish it next Monday.

Does that mean the PR can be merged any time before Monday April 24, or does it need to be merged _on_ the 24th?

deepthi:
> > You can plan to publish it next Monday.
> 
> Does that mean the PR can be merged any time before Monday April 24, or does it need to be merged _on_ the 24th?

If you merge it before Apr 24, it won't appear on the website right away. The first website deploy on or after Apr 24 will publish it.
Since we can't guarantee that there will be a PR merge to trigger a website deploy on that exact day (and we don't want to have to go into netlify and manually trigger a deployment), the best thing to do is to push the green button on Apr 24.

shlomi-noach:
Published URL: https://vitess.io/blog/2023-04-24-schemadiff/

PR #1451, opened 2023-04-24
Expand section on how to do static configuration

dbussink:
This more expands on how to use `caching_sha2_password` and that it's the recommended way forward when using static file auth.

### DIFF
--- a/content/en/docs/17.0/user-guides/configuration-advanced/static-auth.md
+++ b/content/en/docs/17.0/user-guides/configuration-advanced/static-auth.md
@@ -57,15 +57,59 @@ $ mysql -h 127.0.0.1 -u myuser1 -pincorrect_password -e "select 1"
 ERROR 1045 (28000): Access denied for user 'myuser1'
 ```
 
+## Generating passwords
+
+You should always used generated random passwords in static configuration.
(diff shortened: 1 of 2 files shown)


content/en/docs/17.0/user-guides/configuration-advanced/static-auth.md
```diff
@@ -57,15 +57,59 @@ $ mysql -h 127.0.0.1 -u myuser1 -pincorrect_password -e "select 1"
 ERROR 1045 (28000): Access denied for user 'myuser1'
 ```
 
+## Generating passwords
+
+You should always used generated random passwords in static configuration.
+Also when using `caching_sha2_password` or `mysql_native_password` this
+is a must as the formats don't provide any protection against brute force
+attacks if the static
```
frouioui: I think something is missing at the end of this sentence

shlomi-noach: ping @dbussink

dbussink: Fixed!

shlomi-noach (approved):

PR #1454, opened 2023-04-26
Add documentation about grpc-use-static-authentication-callerid flag on vtgate

Phanatic:
Related to vitess PR https://github.com/vitessio/vitess/pull/12961

Depends on: https://github.com/vitessio/vitess/pull/12984


notfelineit:
Do you want to add the new flag to the flags table as well? https://github.com/vitessio/website/blob/prod/content/en/docs/17.0/reference/programs/vtgate.md

deepthi (commented):
The 14.0 and 15.0 changes need to be reverted. This is only relevant for 16.0.2+ (after the original code change is back ported)

content/en/docs/16.0/reference/programs/vtgate.md
```diff
@@ -92,7 +92,8 @@ The following global options apply to `vtgate`:
...
-| --grpc_use_effective_callerid                                    | Boolean               | If set, and SSL is not used, will set the immediate caller id from the effective caller id's principal.                                                                                                                                                                                                                           |                                                                                                                                                                
+| --grpc_use_effective_callerid                                    | Boolean               | If set, and SSL is not used, will set the immediate caller id from the effective caller id's principal.                                                                                                                                                                                                                           |
+| --grpc-use-static-authentication-callerid | boolean | If set, will set the immediate caller id to the username authenticated by the static auth plugin. |
```
deepthi, line 96: @frouioui this flag doesn't exist in 16.0 GA but will go into 16.0.2. What is the right thing to do with the 16.0 docs?

frouioui: @deepthi, that's a good point. My thinking is that the `v16` documentation always refer to the latest `v16` patch release, the one defined in https://vitess.io/docs/releases/#v160. Which is why the sidebar does not say `16.0.0` but simply `16.0`.

frouioui: However, I think we always have this overlap where new documentation for the future patch release is not valid since it has not been released just yet. We never handled that issue in the past so I think it is okay to continue forward with this for now. But if we really wanted to, we would need to have _"release branches"_ for our website repository that we only merge on `prod` once we release the next `vitess` patch release.

deepthi: Yeah we should hold this PR until 16.0.2 is actually out. The cherry-pick PR hasn't even merged yet.

frouioui: @deepthi This PR is now added to our `16.0.2` tracking list, we will merge it only when doing the release.

frouioui (approved):

PR #1456, opened 2023-05-02
Update partial MoveTables vtgate flag name after v15 flag work

mattlord:
The underscore based name was replaced with a dash based one but the docs were not updated.


FancyFane (approved):
Doing a grep search through the code I can see all 7 instances have been replaced with the new dashed flag. LGTM.

PR #1458, opened 2023-05-08
Fix the resharding steps listed in the operator example

GuptaManan100:
### Description

This PR fixes the resharding steps listed in the operator example to also call `Complete` step before stopping the shards. Without this step, we have a workflow still listed in the the `vtctlclient Workflow customer listall` command and Completing it after deleting the shards fails with following error - 
```E0508 17:32:26.610650   52392 main.go:96] E0508 12:02:26.608530 traffic_switcher.go:208] buildTrafficSwitcher failed: node doesn't exist: /vitess/example/global/keyspaces/customer/shards/-/Shard
Reshard Error: rpc error: code = Unknown desc = node doesn't exist: /vitess/example/global/keyspaces/customer/shards/-/Shard
E0508 17:32:26.613887   52392 main.go:105] remote error: rpc error: code = Unknown desc = node doesn't exist: /vitess/example/global/keyspaces/customer/shards/-/Shard
```


rohit-nayak-ps:
We also need to change the latest `17.0` version ...

GuptaManan100:
Fixed it now!

rohit-nayak-ps (approved):

PR #1461, opened 2023-05-09
Replace vtctld web UI callout with VTAdmin one

mattlord:
In the Getting Started / Local Install docs.


notfelineit (approved):
this is great, thank you 🙌

PR #1468, opened 2023-05-11
Add the new VTOrc API to the docs

GuptaManan100:
### Description

This PR adds the documentation for the new VTOrc API endpoint added in https://github.com/vitessio/vitess/pull/12870.


deepthi:
I don't want to document this or support it. Metrics should be added to `/metrics` instead of us inventing new APIs for them.

PR #1478, opened 2023-05-22
doc update: enable settings pool by default

harshit-gangal:
This PR updates the doc for settings pool as it is marked to be "on" for v17 release.


deepthi (commented):
Looks fine, but needs to be copied to v18 docs.

harshit-gangal:
> Looks fine, but needs to be copied to v18 docs.

Made the changes to v18 as well.

frouioui (approved):

PR #1481, opened 2023-05-24
chore: add current release notice to page for SEO boost

thisisobate:
cc: @deepthi @mattlord


mattlord (approved):
Thank you, @thisisobate ! If it's good for SEO then it's good with me. 😄

PR #1495, opened 2023-06-12
Fix typo

tjschuck:


mattlord:
Thanks, @tjschuck ! ❤️ We should make the same correction in the other docs versions as well, at least 18.0. Do you mind making the same change in the other versions?

tjschuck:
@mattlord Updated to fix it everywhere ✌️

rohit-nayak-ps (approved):

Issue #1289, opened 2022-12-15
Improving Website Analytics


harshit-gangal:
Current Analytics on the Vitess Website seems very minimalistic. It gives us good insights about the top pages and demographics. 
But, it can be improved.

It is missing information about the search. About what users are searching for and what they landed on, how much time did they spend on those pages to understand if they landed on the right page or not?

About the flow of pages, users go through. 

This additional information will help us to understand if we should restructure our docs to make information more reachable to the users. 

---

(Edit) Link to the service desk ticket: [CNCFSD-1539](https://cncfservicedesk.atlassian.net/browse/CNCFSD-1539).

harshit-gangal:
Raised the request on cncf service desk.

chalin:
Hi @harshit-gangal. Before deep diving, I have a question to ask so that I can better understand your request.

> Current Analytics on the Vitess Website seems very minimalistic

Just to be clear: are you referring to Universal Analytics (UA) or Google Analytics 4 (GA4)?

/cc @nate-double-u

harshit-gangal:
Hi @chalin
To my knowledge we use GA4

deepthi:
We've been on GA4 since https://github.com/vitessio/website/issues/1098 was fixed.

chalin:
Right, ok, I'm glad that we're on the same page, and all talking about GA4.

> Current Analytics on the Vitess Website seems very minimalistic. ... It is missing information about the search. ...

When I created the GA4 account, I enabled "Enhanced measurement", which includes events for **site search** and **file downloads**, so in that sense it is possible that nothing is missing. Here's an example for file downloads:

> <img width="1141" alt="image" src="https://user-images.githubusercontent.com/4140793/218561604-26c7b665-0cc7-4f54-9834-c5c626de8cc8.png">

It may be more a question of report generation. Feel free to explore the **Reports** and **Explore** sections of the GA4 console for Vitess. Let us know if that suits your needs.

harshit-gangal:
@chalin I tried the `Explore` but I was not able to build a report with the `search term` as a dimension and `page views` 
 as metric.

chalin:
In terms of getting analytics for search from within the site, since Lunr is used, we may indeed need a custom strategy. @thisisobate: might you know how to best hook up Google Analytics to Lunr search? I haven't found anything in yet. Of the top of my head, we could add a search query argument (`?search=<search-term-used>`) to each link in the results pane. WDYT?

[...]

PR #1519, opened 2023-07-01
Track Lunr search queries and search-result clicks

### LINKED ISSUE #1518: Track Lunr search via Google Analytics 4
Originally in the following issue, there was a request to track searched terms on the website:

- #1289

The website uses Lunr to search. Search is incremental, and debounced by 500 ms. While we can sent search terms to GA4 at no more than the debounce rate, this will still result in event tracking for intermediate search terms.

chalin:
- Closes #1518
- Contributes to #1289
- Adds an event handler to send search terms as a `search` event to GA4 with the search term as an attribute, at no more than the debounce rate.
- Adds a `click` event handler to each search-result item. When a user clicks on a search result, a `click` event is sent to GA4 with the destination link and associated search term as attributes.

A note about the **[deploy preview]**: the site preview will allow you to confirm that search is still functional, but it won't generate GA4 events, since analytics is only enabled in production. When `_debug` is enabled (which it was for my original commit, but has since been disabled), you can at least see that the tracking functions get called, like this:

```
trackSearchEvent: searchTerm = installation
trackSearchResultClick: searchTerm = installation, link = https://deploy-preview-1519--vitess.netlify.app/docs/17.0/get-started/local-mac/
```

---

I've locally built a "production" version of the site, and can confirm that GA4 is receiving both `search` and `click` (related to search) events. Notice the `search_term` attribute in both events:

| Event | Event attributes |
|--------|--------|
| **Notice the `search` event with event count of 1:** Screen Shot 2023-07-01 at 11 33 43 | Screen Shot 2023-07-01 at 11 34 04 | 
|  Oops forgot to get a `click`-event screenshot | Screen Shot 2023-07-01 at 11 34 24

/cc @harshit-gangal @thisisobate @nate-double-u 

[deploy preview]: https://deploy-preview-1519--vitess.netlify.app/

### DIFF
--- a/layouts/partials/javascript.html
+++ b/layouts/partials/javascript.html
@@ -26,6 +26,8 @@ $versions.current}}
       "use strict";
       let index;
       let debounceTimer = null;
+      const _debug = true;
+      const debugLog = _debug ? console.log : () => { };
       const currentUrl = window.location.pathname;
       const currentLocation = String(currentUrl).split("/");
       const currentVersion = currentLocation[2];
@@ -139,6 +141,7 @@ $versions.current}}
           }
         });
         search(query);
+        trackSearchEvent(query);
       }
 
       function search(keywords) {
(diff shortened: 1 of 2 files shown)


thisisobate (approved):
LGTM!

harshit-gangal (approved):
Thank you for doing this. Hope to see better analytics now and see how the search is performing for the users.

layouts/partials/javascript.html
```diff
@@ -169,23 +172,51 @@
...
+      }
+
+      function trackSearchResultClick(searchResultURL, searchTerm) {
+        debugLog(`trackSearchResultClick: searchTerm = ${searchTerm}, link = ${searchResultURL}`);
+        if (typeof gtag !== 'function') return;
+        gtag('event', 'click', {
+          event_category: lunrSearchEventCategory,
+          event_label: 'Search-result Click',
+          search_term: searchTerm,
+          value: searchResultURL,
```
harshit-gangal, line 216: from the documentation `value` can be a number only, so it cannot be set to string
https://developers.google.com/tag-platform/devguides/events#gtag.js_1

chalin: For a preliminary analysis, see:

- #1570

PR #1562, opened 2023-08-17
VReplication MoveTables: Initial documentation for the MoveTables foreign key mode

rohit-nayak-ps:
Initial documentation for the foreign key support in MoveTables: implemented in https://github.com/vitessio/vitess/pull/13137.

The only relevant changes below are to do with the `--foreign-key-mode` (working name for the flag, which might change after internal discussions).

The other changes are due to code reformatting.


rohit-nayak-ps:
This is already part of the command reference documentation in v18/v19

PR #1567, opened 2023-08-27
Docs: clarifying which statements are supported by Online DDL

shlomi-noach:
This PR clarifies that Online DDL applies to `CREATE TABLE`, `ALTER TABLE`, `DROP TABLE`, `CREATE VIEW`, `ALTER VIEW`, `DROP VIEW`, and not to other DDL statements such as `TRUNCATE`, `RENAME`, etc.

The doc change is applied to all version `15` .. `18`


GuptaManan100 (approved):
LGTM

PR #1568, opened 2023-08-28
MoveTables: Document --no-routing-rules in v16

rohit-nayak-ps:
Documents flag added in https://github.com/vitessio/vitess/pull/13858. Currently the flag is in v16 only, but will be forward ported to later versions and will be documented separately when that is done.

### DIFF
--- a/content/en/docs/16.0/reference/vreplication/movetables.md
+++ b/content/en/docs/16.0/reference/vreplication/movetables.md
@@ -268,8 +268,8 @@ We caution against against using `EXEC` or `EXEC_IGNORE` for the following reaso
 **default** false
 
 <div class="cmd">
-Do not create routing rules for the tables being moved, when the workflow is created. This implies that you should
-not use global routing or sent traffic to the target keyspace while sending queries to vtgate.
+Do not create routing rules for the tables being moved when the workflow is created. This implies that you should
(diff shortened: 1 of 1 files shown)


content/en/docs/16.0/reference/vreplication/movetables.md
```diff
@@ -262,6 +263,17 @@ We caution against against using `EXEC` or `EXEC_IGNORE` for the following reaso
...
 
+#### --no-routing-rules
+**optional**\
+**default** false
+
+<div class="cmd">
+Do not create routing rules for the tables being moved, when the workflow is created. This implies that you should
+not use global routing or sent traffic to the target keyspace while sending queries to vtgate.
+See https://github.com/vitessio/vitess/pull/13858 and https://github.com/vitessio/vitess/issues/13851 for a use-case
+and more details.
```
mattlord:
```suggestion
Do not create routing rules for the tables being moved when the workflow is created. This implies that you should
not use global routing or send traffic to the target keyspace through a vtgate.
See https://github.com/vitessio/vitess/pull/13858 and https://github.com/vitessio/vitess/issues/13851 for a use-case
and more details.
```

mattlord (approved):

PR #1569, opened 2023-08-28
Correct env.sh path in v16+ local examples docs

### LINKED ISSUE #1523: Vitess 17 Local Install instructions not working - bis.
Bis because of {1} (Vitess 17 Local Install instructions not working).

{1}: https://github.com/vitessio/website/issues/1520

Same context as in {1}, `source ./env.sh` not working in {2}.  My guess is that it should be `source ../common/env.sh`.

{2}: https://vitess.io/docs/17.0/get-started/local/#setup-aliases

```
$ pwd
/home/jgagne/my-vitess-example/examples/local
$ source ./env.sh
bash: ./env.sh: No such file or directory
$ grep source 101_initial_cluster.sh
source ../common/env.sh
$ source ../common/env.sh
$
```

I might submit a PR for this when I am done with exploring Vitess 17.

**Below added after opening the bug...**

Probably similar errors in Vitess 16 {3}, but not in Vitess 15 as it has `env.sh` in `local`.

{3}: https://vitess.io/docs/16.0/get-started/local/#setup-aliases

```
$ grep source vitess-16.0.2-6076fed/examples/local/101_initial_cluster.sh
source ../common/env.sh
$ test -e vitess-16.0.2-6076fed/examples/local/env.sh || echo does not exists
does not exists
```

Also, in {4}, I see `~/my-vitess-example> mysql` but in {5} we `cd ~/my-vitess-example/examples/local`, so it should be  `~/my-vitess-example/examples/local> mysql`.  This error repeats many places in {6} and in the the v16 and v15 versions of this page ({7} and {8}).  I would suggest fixing this by not includinf `pwd` in the prompt for making this less error-prone and easier to understand.

{4}: https://vitess.io/docs/17.0/get-started/local/#start-a-single-keyspace-cluster
{5}: https://vitess.io/docs/17.0/get-started/local/#start-a-single-keyspace-cluster
{6}: https://vitess.io/docs/17.0/get-started/local/
{7}: https://vitess.io/docs/16.0/get-started/local/
{8}: https://vitess.io/docs/15.0/get-started/local/

mattlord:
We also simplify and unify the prompt usage in the commands:
- Use `$ ` as the static prompt
- Only show the prompt when including the output (otherwise the commands are easier to cut and paste w/o it)

Fixes: https://github.com/vitessio/website/issues/1523

### DIFF
--- a/content/en/docs/16.0/get-started/local-brew.md
+++ b/content/en/docs/16.0/get-started/local-brew.md
@@ -215,7 +215,7 @@ rm -rf /usr/local/Cellar/vitess/9.0.0/share/vitess/examples/local/vtdataroot
 For ease-of-use, Vitess provides aliases for `mysql` and `vtctlclient`:
 
 ```bash
-source ./env.sh
+source ../common/env.sh
 ```
 
 Setting up aliases changes `mysql` to always connect to Vitess for your current session. To revert this, type `unalias mysql && unalias vtctlclient` or close your session.
--- a/content/en/docs/16.0/get-started/local-brew.md
+++ b/content/en/docs/16.0/get-started/local-brew.md
@@ -40,9 +40,9 @@ At this point Vitess binaries installed under default Homebrew install location 
 ### Install Node 16.13.0+ (required to run VTAdmin)
 
 ```bash
-$ brew install nvm
-$ nvm install --lts 16.13.0
-$ nvm use 16.13.0
+brew install nvm
+nvm install --lts 16.13.0
+nvm use 16.13.0
 ```
 
 See the [vtadmin README](https://github.com/vitessio/vitess/blob/main/web/vtadmin/README.md) for more details.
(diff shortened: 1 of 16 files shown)


content/en/docs/16.0/get-started/local-brew.md
```diff
@@ -215,7 +219,7 @@ rm -rf /usr/local/Cellar/vitess/9.0.0/share/vitess/examples/local/vtdataroot
 For ease-of-use, Vitess provides aliases for `mysql` and `vtctlclient`:
 
 ```bash
-source ./env.sh
+source ../common/env.sh
```
deepthi, line 222: This was moved to common only in v17, so the v16 and v15 docs should not have any such change.

deepthi: PR: https://github.com/vitessio/vitess/pull/12239

deepthi: Turns out the original PR was back ported to v16.

deepthi (approved):

content/en/docs/16.0/get-started/local-brew.md
mattlord: Yes, here's the backport PR (bot failed): https://github.com/vitessio/vitess/pull/12278

jfg956:
> Only show the prompt when including the output (otherwise the commands are easier to cut and paste w/o it)

I like this, thanks.

PR #1583, opened 2023-09-13
consolidate docs

ajm188:
docs for https://github.com/vitessio/vitess/pull/13959


deepthi (approved):
nit: with this structure Hugo produces a `>` after each program in the left nav whether are not there are any sub-sections. There are additional docs for `vtctl` and `vtctldclient` but not for `vtctld` and `vtorc`. Could be misleading when people are just browsing docs.

ajm188:
okay, i think [707456b](https://github.com/vitessio/website/pull/1583/commits/707456bec40a5abf40a4c021dbd87970cccc22a0) fixes it. i tested in on the https://github.com/vitessio/website/pull/1582 and it had expanders only for `mysqlctl`, `vtctl`, and `vtctldclient` and none of the others

PR #1630, opened 2023-10-31
update FK docs

deepthi:
Rearrange and lightly edit FK docs.


deepthi:
It might be easier to review this by reading the new version of the doc through instead of looking at the diff.

GuptaManan100 (approved):
lgtm

PR #1652, opened 2023-12-20
Update docs for release 16.0.7, 17.0.5, and 18.0.2

mattlord:


frouioui (approved):
LGTM besides the length of the SHA that should be 7 instead of 10

harshit-gangal (approved):

PR #1654, opened 2023-12-20
Highlight search keywords on search result

### LINKED ISSUE #1372: Highlight the search keyword on the search result
When a user searches for a keyword, the results do not give any clue about what the user is searching for. 
To improve this, we need a way to dynamically create a summary that will highlight the search keyword on the results page.

Maniktherana:
closes #1372 

Preview:
Screenshot 2023-12-20 at 10 41 55 PM

### DIFF
--- a/layouts/partials/javascript.html
+++ b/layouts/partials/javascript.html
@@ -182,16 +182,22 @@ $versions.current}}
             const title = searchResultItem.querySelector('a.title');
             title.setAttribute('href', doc.href);
             title.textContent = doc.title;
+            title.innerHTML = highlightSearchTerm(doc.title, keywords);
 
             const subtitle = searchResultItem.querySelector('a.subtitle');
             subtitle.setAttribute('href', doc.href);
             subtitle.textContent = doc.summary;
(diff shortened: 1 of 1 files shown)


layouts/partials/javascript.html
```diff
@@ -194,6 +196,20 @@
         }
       }
 
+      function highlightSearchTerm(content, searchTerm) {
+        const keywords = searchTerm.trim().split(/\s+/); // Split keywords by empty space
```
mattlord: Just noting that I hope that at some point we support searching for quoted strings. It makes searching for some things challenging. But today splitting these up into individual tokens by whitespace like this matches the behavior.

mattlord (approved):
Thanks, @Maniktherana! ❤️

It seems to work as expected here: https://deploy-preview-1654--vitess.netlify.app/

@chalin would you mind taking a look as well if you have a moment? This is very much not my area of expertise. 🙂

frouioui (approved):
This is a good feature! Thanks a bunch @Maniktherana. I will merge it now as it has been first reviewed a month ago. We can re-open and revert if @chalin has feedback

PR #1658, opened 2024-01-04
Enhance the getting started documentation with proper versions

frouioui:
This PR updates the `Get Started` guides even more by adding the proper `git checkout` command to ensure users are using the version of Vitess that matches the documentation.

Moreover, the Homebrew install page is removed. Our tap only serves our latest version, to avoid having to maintain our own tap or more automation, @mattlord and I have decided to remove the documentation related to Homebrew.


mattlord (approved):
❤️ Thanks, again @frouioui !

PR #1683, opened 2024-02-08
add chapter about vindex hints

systay:
Should only be merged when https://github.com/vitessio/vitess/pull/15172 is merged


GuptaManan100 (approved):
Love it!

PR #1684, opened 2024-02-08
rm docs about --warn-non-atomic-commit

maxenglander:
There is no such flag. In v19, this warning behavior is on and cannot be disabled.


frouioui (commented):
I am not seeing such flag in Vitess' main as well. We should perhaps remove the documentation for v20 too

maxenglander:
@frouioui thanks, good catch, fixed.

mattlord (approved):
Thanks, @maxenglander !

PR #1710, opened 2024-03-21
fix(supported-databases): update 8.0 to 8.0.* to make it clear to people any version of 8.0.* should work.

dbmurphy:
fix(supported-databases): update 8.0 to 8.0.* to make it clear to people any version of 8.0.* should work.

### DIFF
--- a/content/en/docs/19.0/overview/supported-databases.md
+++ b/content/en/docs/19.0/overview/supported-databases.md
@@ -12,7 +12,9 @@ The [VTGate](../../concepts/vtgate/) proxy server advertises its version as MySQ
 
 ### MySQL 8.0
 
-Vitess supports the core features of MySQL 8.0, with [some limitations](../../reference/compatibility/mysql-compatibility/). Vitess also supports [Percona Server for MySQL](https://www.percona.com/software/mysql-database/percona-server) 8.0.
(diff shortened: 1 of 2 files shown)


deepthi:
I know this was discussed on slack, and I'm kind of with Matt here. MySQL 8.0 is understood to mean to include all minor versions, so I don't see the need to add a `*`.

dbmurphy:
> I know this was discussed on slack, and I'm kind of with Matt here. MySQL 8.0 is understood to mean to include all minor versions, so I don't see the need to add a `*`.

If that's how you see it, I can't help, but I was clear that multiple people in our organization see it differently, so it is a problem for people who are not as close to the product as we are.  I think where @mattlord  and I landed, it doesn't hurt anything to add the `*` to be more precise for others, especially in business and security roles.  

The comments were along lines such as "It's ambiguous and unclear if there may be any 8.0.* version or a specific one listed elsewhere" or "Can we only use the 8.0 that they used in the example, which did not change from v18.0 to v19.0"?

deepthi:
> multiple people in our organization see it differently, so it is a problem for people who are not as close to the product as we are.

OK, this is a valid argument. We want docs to be clear to people who might be coming new to them, so let's go ahead and merge this.

deepthi (approved):

PR #1729, opened 2024-04-22
Fixes for the mac documentation

bddicken:
A few small fixes for the Mac install page:

* Use regular quotes instead of smart quotes
* Remove two non-displaying characters
* use `${HOME}` instead of `/Users/jason`
* Remove `$`

### DIFF
--- a/content/en/docs/17.0/get-started/local-mac.md
+++ b/content/en/docs/17.0/get-started/local-mac.md
@@ -45,14 +45,14 @@ See the [vtadmin README](https://github.com/vitessio/vitess/blob/main/web/vtadmi
 With the tools you’ve just installed via brew, you will next update your PATH variable so your shell knows where to find the binaries:
 
 ```sh
-echo "export PATH=${PATH}:/opt/homebrew/opt/mysql-client/bin:/opt/homebrew/opt/mysql/bin:~/Github/vitess/bin:${HOME}/go/bin:/opt/homebrew/bin" >> ~/.zshrc
+echo "export PATH=${PATH}:/opt/homebrew/opt/mysql-client/bin:/opt/homebrew/opt/mysql/bin:${HOME}/go/bin:/opt/homebrew/bin" >> ~/.zshrc
 source ~/.zshrc
 ```
 
 If you’re using bash for your shell you’ll have to update the paths in `.bash_profile` or `.bashrc` instead. Mac does not read `.bashrc` by default:
 
 ```sh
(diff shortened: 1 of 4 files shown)


bddicken:
Also, this PR does not include changes to the docs in `content/en/docs/archive/` but if you'd like me to update those as well just let me know.

content/en/docs/17.0/get-started/local-mac.md
```diff
@@ -45,14 +45,14 @@ See the [vtadmin README](https://github.com/vitessio/vitess/blob/main/web/vtadmi
 With the tools you’ve just installed via brew, you will next update your PATH variable so your shell knows where to find the binaries:
 
 ```sh
-echo “export PATH=${PATH}:/opt/homebrew/opt/mysql-client/bin:/opt/homebrew/opt/mysql/bin:~/Github/vitess/bin:/Users/jason/go/bin:​​/opt/homebrew/bin” >> ~/.zshrc
+echo "export PATH=${PATH}:/opt/homebrew/opt/mysql-client/bin:/opt/homebrew/opt/mysql/bin:~/Github/vitess/bin:${HOME}/go/bin:/opt/homebrew/bin" >> ~/.zshrc
```
mattlord: We should also replace this: `~/Github/vitess/bin` with `${VTROOT}/bin` here and below

mattlord (approved):
Thanks, @bddicken ! I noted one additional change that we should make since we're here. No worries about the archived pages.

content/en/docs/17.0/get-started/local-mac.md
bddicken: Perhaps I'm missing something here, but I'm not sure we'd want to use `${VTROOT}` here, since this is just going into `.zshrc`. That would not have been set yet, if someone were just following the instructions on this docs page, right?

Or did you mean, just use `${HOME}`?

mattlord: `~/Github/vitess/bin` would only be useful if `~/Github/vitess` is where you cloned the vitess repo (which is what $VTROOT gets set to) and you've at some point run `make build`. The Makefile does this: https://github.com/vitessio/vitess/blob/a89b2cb2d11686a49e1c31886126076689772c7e/Makefile#L68-L74

The `build.env` file also has this: https://github.com/vitessio/vitess/blob/main/build.env#L27

You could also remove `~/Github/vitess/bin` if you prefer. It would be entirely useless to anyone that didn't happen to use that structure (I don't).

bddicken: Ohh, right. I think for this step in particular, it makes sense to just get rid of that part of the path. This comes before even cloning the vitess repo. Also, a few steps later, the instructions solve this problem with:

```
$ cd bin
$ echo "$(printf 'export PATH="${PATH}:'; echo "$(pwd)\"")" >> ~/.zshrc
$ source ~/.zshrc
```

PR #1736, opened 2024-05-01
Documentation fixes

bddicken:
I've been going through vitess docs / building / walkthroughs and noticed a few small issues in the docs:

- In the mac docs, do not need to explicitly execute `source ../common/env.sh` before `./101_initial_cluster.sh`. `./101_initial_cluster.sh` does that within the script!

- On `Ubuntu 24.04 LTS`, `sudo apt install etcd` does not work. I tried several things to get it to work, and ultimately settled on installing both `etcd-server` and `etcd-client`. Perhaps I'm missing something, so please let me know if I'm off here.

### DIFF
--- a/content/en/docs/17.0/get-started/local.md
+++ b/content/en/docs/17.0/get-started/local.md
@@ -16,7 +16,7 @@ Vitess supports the databases listed [here](../../overview/supported-databases/)
 
 ```sh
 # Ubuntu based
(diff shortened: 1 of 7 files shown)


bddicken:
@mattlord All the comments you mentioned should be fixed for this PR. lmk what you think.

mattlord (approved):
Thanks, @bddicken !

Issue #1751, opened 2024-05-19
Add CODE_OF_CONDUCT.md to Repository


Ayush9026:
## Description:

**Overview :**
To foster a welcoming and inclusive community, I propose adding a CODE_OF_CONDUCT.md file to the Vitess website repository. This file will outline expected behavior and guidelines for maintaining a positive environment.

**Objectives :**

1. Promote Inclusivity: Establish clear behavior guidelines.
2. Provide Support: Offer a framework for reporting incidents.
3. Foster Community Spirit: Encourage respect and collaboration.

**Conclusion**

Adding a CODE_OF_CONDUCT.md file will ensure a positive and inclusive environment for all Vitess community members.

deepthi:
We already have a [code of conduct in the main repo](https://github.com/vitessio/vitess/blob/main/CODE_OF_CONDUCT.md). Why would we duplicate it here?

Ayush9026:
Thank you mam for pointing that out.

PR #1752, opened 2024-05-19
Add CODE_OF_CONDUCT.md file

Ayush9026:
This PR solves the issue https://github.com/vitessio/website/issues/1751 
To foster a welcoming and inclusive community, I propose adding a CODE_OF_CONDUCT.md file to the Vitess website repository. This file will outline expected behavior and guidelines for maintaining a positive environment.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,5 @@
+## Vitess Community Code of Conduct
+
+Vitess follows the CNCF Code of Conduct.
+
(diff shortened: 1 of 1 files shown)


deepthi:
It doesn't make sense to create another file here. If at all, we should update the file in the main repo, and link it from the Contributing page.

Ayush9026:
You're right, it doesn't make sense to create another file here, ma'am. If at all, we should update the file in the main repo and link to it from the Contributing page. I will do this.

PR #1760, opened 2024-05-24
Adding in context about noblob and schema changes.

hallaroo:
Adding in additional context in the location noted by @deepthi. Thanks!


deepthi (commented):
Thanks @hallaroo!
Do you mind adding this to all versions of the docs? 18,19,20.
And remove the .DS_Store file.

content/en/docs/17.0/user-guides/schema-changes/managed-online-schema-changes.md
```diff
@@ -241,3 +241,8 @@ The tables are kept for 24 hours after migration completion. Vitess automaticall
 
 Artifact tables are identifiable via `artifacts` column in a `SHOW VITESS_MIGRATION ...` command. You should generally not touch these tables. It's possible to `DROP` those tables with `direct` DDL strategy. Note that dropping tables in production can be risky and lock down your database for a substantial period of time. Dropping artifact tables also makes the migrations impossible to [revert](../revertible-migrations/).
 
+## Noblob
+
+The noblob binlog row image is supported by the MoveTables and Reshard VReplication workflows. If noblob is used it won't affect schema changes as long as blobs are not part of the PRIMARY KEY, in most vreplication workflows. 
+
+**NOTE** BLOB should never be part of a PRIMARY KEY and should also not be part of the Online DDL migration key.
```
deepthi: @shlomi-noach is this clear enough? or should we elaborate on what we mean by Online DDL migration key?

shlomi-noach: It's clear enough, although the rules are even stricter: in MySQL you can only have a blob in your `PRIMARY KEY` when using prefix length index. Thus, this definition is invalid:

```sql
create table t (
b blob primary key
);
```

You _can_ create a `PRIMARY KEY` using a blob prefix:
```sql
create table t (
b blob not null,
primary key (b(20))
);
```

However, vreplication does not support index prefixes for the iteration key (primary key). And that's the even-more-basic limitation to why you would not use blogs in primary key / iteration key in the first place. That's irrespective of `noblob`. That limitation then plays nicely into the use cases that require `noblob`.

So I think a bit of rewording to clarify that blobs are never supported by vreplication as part of the `PRIMARY KEY` or the migration key.

Re: should we elaborate on what we mean by Online DDL migration key, we can just link to this doc page: https://vitess.io/docs/archive/15.0/reference/vreplication/internal/keys/

content/en/docs/17.0/user-guides/schema-changes/managed-online-schema-changes.md
```diff
@@ -241,3 +241,8 @@ The tables are kept for 24 hours after migration completion. Vitess automaticall
 
 Artifact tables are identifiable via `artifacts` column in a `SHOW VITESS_MIGRATION ...` command. You should generally not touch these tables. It's possible to `DROP` those tables with `direct` DDL strategy. Note that dropping tables in production can be risky and lock down your database for a substantial period of time. Dropping artifact tables also makes the migrations impossible to [revert](../revertible-migrations/).
 
+## Noblob
+
+The noblob binlog row image is supported by the MoveTables and Reshard VReplication workflows. If noblob is used it won't affect schema changes, as long as blobs are not part of the PRIMARY KEY in vreplication workflows. 
+
+**NOTE** BLOB should never be part of a PRIMARY KEY and should also not be part of the Online DDL [migration key](https://vitess.io/docs/archive/15.0/reference/vreplication/internal/keys/). This is becuse vreplication does not support index prefixes for the PRIMARY KEY.
```
deepthi: This link needs to be relative so that each version's doc points to the correct version of this page. Also because we don't really want to link to archived docs from live ones.

deepthi: @shlomi-noach can you suggest a re-wording for ` vreplication does not support index prefixes for the PRIMARY KEY.`? that is not exactly what we want to say.

deepthi: Maybe just `vreplication does not support such usage`? or `vreplication does not support it`?

shlomi-noach:
```suggestion
The `noblob` binlog row image is supported by the MoveTables and Reshard VReplication workflows.

Note: vreplication does not support blobs as part of the `PRIMARY KEY` in `MoveTable` operations, or as part of the migration key in Online DDL operations. 
```

shlomi-noach (approved):
Made a suggestion to simplify the `noblob` discussion.

deepthi (approved):

PR #1781, opened 2024-07-04
Remove local-docker-guide from v21

kirtanchandak:
- removed local docker guide from v21


frouioui (commented):
LGTM, it's debatable whether we want to do this on all supported docs version (18, 19, 20, 21) or just 21. I don't have strong opinions, but if we decide to do so, we shoudl backport https://github.com/vitessio/vitess/pull/16335 to the corresponding branches.

kirtanchandak:
> LGTM, it's debatable whether we want to do this on all supported docs version (18, 19, 20, 21) or just 21. I don't have strong opinions, but if we decide to do so, we shoudl backport [vitessio/vitess#16335](https://github.com/vitessio/vitess/pull/16335) to the corresponding branches.

@deepthi what should we do?

deepthi:
> LGTM, it's debatable whether we want to do this on all supported docs version (18, 19, 20, 21) or just 21. I don't have strong opinions, but if we decide to do so, we shoudl backport [vitessio/vitess#16335](https://github.com/vitessio/vitess/pull/16335) to the corresponding branches.

Let's do this on just v21 for now. If at all, we may do it on v20 as well, but no need to do any older releases. And even if we decide to change the docs on older releases, there's no need to backport the Makefile change, it doesn't hurt anything to leave that there if there is no reference to it in the docs.

deepthi (approved):

PR #1791, opened 2024-07-18
Update zoom links

frouioui:
We recently changed the zoom link for the vitess monthly meeting. Instead of trying to keep this current, we link to the CNCF calendar which will always be up-to-date.


deepthi (commented):
Instead of linking directly to zoom here, let us link to the cncf calendar entry. That way we won't need to edit these pages again. 
We can probably delete the dialin.md page altogether after making that change.

frouioui:
> Instead of linking directly to zoom here, let us link to the cncf calendar entry. That way we won't need to edit these pages again. We can probably delete the dialin.md page altogether after making that change.

@deepthi, done via 5aa46ff

deepthi (approved):

Issue #1763, opened 2024-06-03
Fix overflow issue in video container


jayesh9747:
The video container is causing an overflow issue on smaller screens, resulting in part of the content being cut off and not properly displayed.

![image](https://github.com/vitessio/website/assets/123955234/00663c37-5c20-42cb-a550-c9854b9f995c)
![image](https://github.com/vitessio/website/assets/123955234/701b7dde-94b2-43fb-8abe-1ac7b66aae77)

thisisobate:
Issue already fixed [here](https://github.com/vitessio/website/pull/1794). I think we should close this one.
cc: @deepthi @frouioui

frouioui:
Closing this, I can confirm that it was fixed.

PR #1794, opened 2024-07-23
updating video container is-video

nate-double-u:
contributes to https://github.com/vitessio/website/issues/1763

updates video container `is-video` so it doesn't overflow when the window is resized.

### DIFF
--- a/assets/sass/style.sass
+++ b/assets/sass/style.sass
@@ -142,7 +142,12 @@ figure
     flex: 1
 
 .is-video
-  width: 100%
+  max-width: 100%
+  overflow: hidden
+  iframe,
+  video 
+    width: 100%
+    height: 100%
 
 .column
   &.is-thin


deepthi:
I looked at the preview. While it doesn't overflow any more, it does get cut-off on both left and right. Is there a way to get the container resized on mobile (to be smaller) so that we avoid this?

nate-double-u:
interesting, i see that now too. I think it's just the title-card that doesn't resize correctly, the video still plays right.

I'll see what I can do.

chalin (commented):
You'll want to drop the semicolons, btw.

chalin:
@nate-double-u - is `bulma` due for an upgrade? Might that upgrade resolve the issue this PR is trying to solve?

nate-double-u:
I tried updating Bulma locally (the site is only a dot version behind), and it didn't help. I'm not sure why the title card isn't scaling with the rest of the video, but the video behaves more correctly now.

That is to say in my testing it's only the title card that gets cut off, not the video. (Please correct me if I'm wrong here, I've admittedly only tested on Firefox and mobile)

deepthi:
> I tried updating Bulma locally (the site is only a dot version behind), and it didn't help. I'm not sure why the title card isn't scaling with the rest of the video, but the video behaves more correctly now.
> 
> That is to say in my testing it's only the title card that gets cut off, not the video. (Please correct me if I'm wrong here, I've admittedly only tested on Firefox and mobile)

That is also what I'm seeing on iOS/Safari. It is at least better than before, if we want to call this "good enough for now".

IMG_9224

deepthi:
@nate-double-u if you want to merge this, you will need to remove the semi-colons. I can approve after that.

nate-double-u:
> @nate-double-u if you want to merge this, you will need to remove the semi-colons. I can approve after that.

Removed

deepthi (approved):

PR #1802, opened 2024-07-31
Add edit button in upper right of docs

rbranson:
It's easy to miss that there's an easy-to-access edit workflow link at the bottom of the page, so this change adds the same link at the top right corner of the documentation for more visibility.


notfelineit (approved):

deepthi (approved):
Nice work!

Issue #1738, opened 2024-05-08
Review and improve Getting Started guides


deepthi:
We have had the Getting Started guides for a while. It is time to re-run them and make sure that everything still works as expected.

asr2003:
@deepthi Any updates on this issue? I'd like to work on this issue

frouioui:
Related to https://github.com/vitessio/website/issues/1798. We are proposing an LFX project to work on this for the 3rd term of 2024.

asr2003:
Thanks @frouioui for the info! I thought it is another issue apart from it :)

asr2003:
Any guidance on how to get started will be greatful. Maybe a pretest or technical design doc to know more about vitess

deepthi:
This particular issue is already being worked on. Please note that It is not part of the scope of LFX term 3, it's just related. PRs are still welcome for things I might miss as improvements.

PR #1803, opened 2024-08-01
Improve code block formatting in Getting Started guide - Local Install

mhmohona:
I have noticed the inconsistency of the code block formatting which confused me as a new user while going through the [local installation](https://vitess.io/docs/20.0/get-started/local/) guide.  

Screenshot from 2024-08-01 09-45-20

Related Issue: #1738

### DIFF
--- a/content/en/docs/21.0/get-started/local.md
+++ b/content/en/docs/21.0/get-started/local.md
@@ -84,15 +84,15 @@ sudo setenforce 0
 
 ## Install Vitess
 
-Download the [latest binary release](https://github.com/vitessio/vitess/releases) for Vitess on Linux. For example with Vitess 19:
+Download the [latest binary release](https://github.com/vitessio/vitess/releases) for Vitess on Linux. For example:
 
 **Notes:**
 
 * Ubuntu is the only fully supported OS, for another OS please [build Vitess by yourself](/docs/contributing) or use the Docker images.
 
 ```sh
-version=20.0.1
-file=vitess-${version}-003c441.tar.gz
+version=20.0.0-rc2
+file=vitess-${version}-4af99b5.tar.gz
 wget https://github.com/vitessio/vitess/releases/download/v${version}/${file}
 tar -xzf ${file}
 cd ${file/.tar.gz/}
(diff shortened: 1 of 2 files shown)


mattlord (approved):
Thanks @mhmohona ! ❤️

mattlord (changes requested):
Thanks, @mhmohona ! I confirmed that it all looks good here: https://deploy-preview-1803--vitess.netlify.app/docs/20.0/get-started/local/

The only thing we need to do is make the same changes in the 21.0 docs, otherwise the changes in v20 will eventually be lost when that version is EOL'd.

mhmohona:
Thank you for pointing that @mattlord! Updating it right away. 

Another thing I think we can improve is, currently both code block and output block is in dark color. As new user, the output block really helped me, but I think if the output block, like the one in [Start a Single Keyspace Cluster](https://vitess.io/docs/20.0/get-started/local/#start-a-single-keyspace-cluster) section was in different color, it would be more understandable for user to differentiate. What do you think @mattlord?

mattlord:
> Another thing I think we can improve is, currently both code block and output block is in dark color. As new user, the output block really helped me, but I think if the output block, like the one in [Start a Single Keyspace Cluster](https://vitess.io/docs/20.0/get-started/local/#start-a-single-keyspace-cluster) section was in different color, it would be more understandable for user to differentiate. What do you think @mattlord?

Yeah, I'm certainly open to that. I think the only consideration is that the contrast is relatively high for people with visual impairment.

content/en/docs/21.0/get-started/local.md
```diff
@@ -74,22 +76,23 @@ sudo aa-status | grep mysqld
...
+
 ```sh
 # CentOS
 sudo setenforce 0
 ```
 
 ## Install Vitess
 
-Download the [latest binary release](https://github.com/vitessio/vitess/releases) for Vitess on Linux. For example:
+Download the [latest binary release](https://github.com/vitessio/vitess/releases) for Vitess on Linux. For example with Vitess 19:
```
mattlord: This should say Vitess 20.

mattlord (approved):

PR #1805, opened 2024-08-02
Unifying common build issues on MacOS, Centos and Ubuntu

kirtanchandak:
- added missing issues in centos

### DIFF
--- a/content/en/docs/contributing/build-on-centos.md
+++ b/content/en/docs/contributing/build-on-centos.md
@@ -142,12 +142,6 @@ mysqld: [ERROR] Fatal error in defaults handling. Program aborted!
 E1027 18:28:23.464780   19483 mysqld.go:734] mysqld --initialize-insecure failed: /usr/sbin/mysqld: exit status 1, output: mysqld: [ERROR] Failed to open required defaults file: /home/morgo/vitess/vtdataroot/vt_0000000101/my.cnf
 mysqld: [ERROR] Fatal error in defaults handling. Program aborted!
 ```
-The following command disables the AppArmor profile for `mysqld`:
-
-```
-sudo ln -s /etc/apparmor.d/usr.sbin.mysqld /etc/apparmor.d/disable/
-sudo apparmor_parser -R /etc/apparmor.d/usr.sbin.mysqld
-```
 
 The following command should now return an empty result:
 ```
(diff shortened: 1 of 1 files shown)


kirtanchandak:
@deepthi should I also add these in the MacOS guide, will these solutions work there too?

content/en/docs/contributing/build-on-centos.md
```diff
@@ -142,4 +142,41 @@ mysqld: [ERROR] Fatal error in defaults handling. Program aborted!
 E1027 18:28:23.464780   19483 mysqld.go:734] mysqld --initialize-insecure failed: /usr/sbin/mysqld: exit status 1, output: mysqld: [ERROR] Failed to open required defaults file: /home/morgo/vitess/vtdataroot/vt_0000000101/my.cnf
 mysqld: [ERROR] Fatal error in defaults handling. Program aborted!
 ```
+The following command disables the AppArmor profile for `mysqld`:
```
deepthi: I don't believe AppArmor is default on CentOS, so this section is not needed. The counterpart on CentOS is SELinux which is already covered in the guide.

content/en/docs/contributing/build-on-centos.md
```diff
@@ -143,3 +143,34 @@ E1027 18:28:23.464780   19483 mysqld.go:734] mysqld --initialize-insecure failed
...
+
+```
+for process in `pgrep -f '(vtdataroot|VTDATAROOT)'`; do 
+ kill -9 $process
+done;
+
+export VTDATAROOT=/tmp/vtdataroot
+./101_initial_cluster.sh
+```
+
```
deepthi: This entire section is about AppArmor, it needs to be deleted. It wasn't enough to delete one line.

content/en/docs/contributing/build-on-centos.md
```diff
@@ -128,18 +128,18 @@ Error:  105: Key already exists (/vitess/zone1) [6]
 Error:  105: Key already exists (/vitess/global) [6]
 ```
 
-### MySQL Fails to Initialize
-
-This error is most likely the result of SELinux enabled:
-
+### Mysqlctl Fails to Start 
+This error is because some tests need `VTROOT`, but it is missing. You can manually run `source dev.env` to solve this:
 ```
```
deepthi, line 133: This still isn't correct. You seem to have replaced an existing section with a new one instead of adding the new one at the end.

deepthi:
> should I also add these in the MacOS guide, will these solutions work there too?

No need to add anything there, AppArmor and SELinux are not relevant on MacOS.

PR #1809, opened 2024-08-09
Fixed latest release doc

GrahamCampbell:


content/en/docs/releases/_index.md
```diff
@@ -15,7 +15,7 @@ Each major release is maintained for 1 year.
 
 ## Release History
 
-> The latest and current vitess release is v17.0
+> The latest and current vitess release is v20.0
```
mattlord, line 18: I think that we could also remove this line, but it should at least be correct 😄

mattlord (approved):
Thanks, @GrahamCampbell ! ❤️

PR #1812, opened 2024-08-12
Add architecture diagram for VReplication

mhmohona:
Added architecture diagram for VReplication and update some formatting for consistency.


mattlord (changes requested):
Thank you for contributing, @mhmohona ! I'm afraid, however, that this isn't really accurate or very helpful IMO.

mhmohona:
> Thank you for contributing, @mhmohona ! I'm afraid, however, that this isn't really accurate or very helpful IMO.

Thank you for your kind response @mattlord. By saying it's not helpful, do you mean about the diagram?
Regarding accuracy, would you please point out where it can be improved? Or the code where I can look?

Issue #1787, opened 2024-07-12
Incorrect page forward


bddicken:
When visiting this URL:

https://vitess.io/docs/reference/features/vtgate-buffering/

I expect it to forward me to the associated page for the current version of Vitess. In this case, the v20 page:

https://vitess.io/docs/20.0/reference/features/vtgate-buffering/

However, it takes me here instead:

https://vitess.io/docs/archive/17.0/user-guides/configuration-advanced/buffering-scenarios/

This seems incorrect.

anshikavashistha:
@bddicken I would like to work on this issue

anshikavashistha:
Could you please assign it to me ?
@bddicken

shailpujan88:
@bddicken I would like to work on this issue. Can you please assign it to me?

mattlord:
@anshikavashistha are you still working on this?

shailpujan88:
Hey @mattlord Can you please assign it to me?

anshikavashistha:
Yes I am 
@mattlord

anshikavashistha:
PTAL @bddicken

frouioui:
Fixed by #1834

frouioui:
Works as expected after the new deploy:

https://vitess.io/docs/reference/features/vtgate-buffering/

Redirects to

https://vitess.io/docs/20.0/reference/features/vtgate-buffering/

PR #1814, opened 2024-08-14
Fix redirect logic and link references for buffering scenarios documentation

anshikavashistha:
This PR addresses the issue [#1787](https://github.com/vitessio/website/issues/1787) where links without an explicit version redirect to the incorrect version and page. Specifically, the link for VTGate Buffering is incorrectly redirecting to the v17 archive page instead of the current v20 page. The redirect logic needs to be updated to point to the correct version and page.

**Changes Made**
- Updated Redirect Logic: Adjusted the redirect logic to ensure that links without an explicit version correctly redirect to the current version (20.0) and the appropriate page.

- Fixed Link References: Changed all relative links in the v17 page to ensure they properly redirect to the current version’s page. This avoids linking to outdated content and ensures users land on the correct page.

**Testing**
vt buffer
vt buffer 2

### DIFF
--- a/content/en/docs/archive/17.0/user-guides/configuration-advanced/buffering-scenarios.md
+++ b/content/en/docs/archive/17.0/user-guides/configuration-advanced/buffering-scenarios.md
@@ -6,7 +6,7 @@ aliases: ['/docs/reference/features/vtgate-buffering',
 ---
 
 For documentation on buffering behaviors please see
-[VTGate Buffering](../../../reference/features/vtgate-buffering/).
+[VTGate Buffering](https://vitess.io/docs/20.0/reference/features/vtgate-buffering/).
 In this guide we are going to go through a few scenarios involving buffering to
 see the practical behaviors. There are several parameters to tune for buffering
 so, we will be using a python utility [gateslap](https://github.com/planetscale/gateslap)
--- a/content/en/docs/archive/17.0/user-guides/configuration-advanced/buffering-scenarios.md
+++ b/content/en/docs/archive/17.0/user-guides/configuration-advanced/buffering-scenarios.md
@@ -6,7 +6,7 @@ aliases: ['/docs/reference/features/vtgate-buffering',
 ---
 
 For documentation on buffering behaviors please see
-[VTGate Buffering](https://vitess.io/docs/20.0/reference/features/vtgate-buffering/).
+[VTGate Buffering]( /docs/20.0/reference/features/vtgate-buffering ).
 In this guide we are going to go through a few scenarios involving buffering to
 see the practical behaviors. There are several parameters to tune for buffering
 so, we will be using a python utility [gateslap](https://github.com/planetscale/gateslap)
--- a/layouts/index.redirects
+++ b/layouts/index.redirects
@@ -66,6 +66,14 @@
 /docs/11.0/                              /docs/archive/11.0/
 /docs/11.0/*                             /docs/archive/11.0/:splat
 
+# Redirect old version documentation to current version
+/docs/archive/17.0/reference/features/vtgate-buffering     /docs/20.0/reference/features/vtgate-buffering 301
+     
+
+# Redirect old version directories to current version
+/docs/17.0/                              /docs/20.0/ 301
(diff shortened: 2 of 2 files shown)


anshikavashistha:
@mattlord PTAL.I have adjusted the redirect logic to ensure that links without an explicit version correctly redirect to the current version (20.0) and the appropriate page.

mattlord:
> @mattlord PTAL.I have adjusted the redirect logic to ensure that links without an explicit version correctly redirect to the current version (20.0) and the appropriate page.

I appreciate you working on this! I'm afraid that this still isn't right, however. This adds some explicit overrides that will cause their own problems and won't solve the underlying issue — which I'm assuming we'd have again when v21 becomes the current version.

anshikavashistha:
> > @mattlord PTAL.I have adjusted the redirect logic to ensure that links without an explicit version correctly redirect to the current version (20.0) and the appropriate page.
> 
> I appreciate you working on this! I'm afraid that this still isn't right, however. This adds some explicit overrides that will cause their own problems and won't solve the underlying issue — which I'm assuming we'd have again when v21 becomes the current version.

Alright @mattlord What could be the best way to overcome this problem in the explicit overriding?

mattlord:
> > > @mattlord PTAL.I have adjusted the redirect logic to ensure that links without an explicit version correctly redirect to the current version (20.0) and the appropriate page.
> > 
> > 
> > I appreciate you working on this! I'm afraid that this still isn't right, however. This adds some explicit overrides that will cause their own problems and won't solve the underlying issue — which I'm assuming we'd have again when v21 becomes the current version.
> 
> Alright @mattlord What could be the best way to overcome this problem in the explicit overriding?

I'm not sure where this is coming from and why it's happening. I've already spent some time investigating it when the issue was first created. If I knew how I would have fixed it. 😄

anshikavashistha:
> > > > @mattlord PTAL.I have adjusted the redirect logic to ensure that links without an explicit version correctly redirect to the current version (20.0) and the appropriate page.
> > > 
> > > 
> > > I appreciate you working on this! I'm afraid that this still isn't right, however. This adds some explicit overrides that will cause their own problems and won't solve the underlying issue — which I'm assuming we'd have again when v21 becomes the current version.
> > 
> > 
> > Alright @mattlord What could be the best way to overcome this problem in the explicit overriding?
> 
> I'm not sure where this is coming from and why it's happening. I've already spent some time investigating it when the issue was first created. If I knew how I would have fixed it. 😄

Alright @mattlord.May be this will work a sort of 
```bash 
# Dynamic redirect logic for current and next versions
/docs/current /docs/{{ $current }}/ 301
/docs/current/* /docs/{{ $current }}/:splat 301

# Redirects for old versions should point to the current version
/docs/17.0/* /docs/{{ $current }}/:splat 301
```

mattlord:
> Alright @mattlord.May be this will work a sort of
> 
> ```shell
> # Dynamic redirect logic for current and next versions
> /docs/current /docs/{{ $current }}/ 301
> /docs/current/* /docs/{{ $current }}/:splat 301
> 
> # Redirects for old versions should point to the current version
> /docs/17.0/* /docs/{{ $current }}/:splat 301
> ```

This breaks the entire model. This issue is somehow specific to that ONE page (AFAIK so far anyway). It's better to leave that one broken than break the model for all pages.

anshikavashistha:
> > Alright @mattlord.May be this will work a sort of
> > ```shell
> > # Dynamic redirect logic for current and next versions
> > /docs/current /docs/{{ $current }}/ 301
> > /docs/current/* /docs/{{ $current }}/:splat 301
> > 
> > # Redirects for old versions should point to the current version
> > /docs/17.0/* /docs/{{ $current }}/:splat 301
> > ```
> 
> This breaks the entire model. This issue is somehow specific to that ONE page (AFAIK so far anyway). It's better to leave that one broken than break the model for all pages.

@mattlord Should I stop working on this issue right now?

mattlord:
@anshikavashistha that's up to you. I just want to be clear on what the problem is and how we should fix it.

PR #1825, opened 2024-08-20
add short blog about recursive CTEs

systay:
Preview: https://deploy-preview-1825--vitess.netlify.app/blog/2024-08-20-recursive-cte/


shlomi-noach (approved):

mattlord (approved):
❤️

deepthi:
We agreed offline that we need to call out that this feature is currently only available on main, and is expected to land in v21. I added the `do-not-merge` label so that we don't merge it until that change has been made.

frouioui (approved):
Taking into account what @deepthi commented and the length of the blog post, I wonder if this bit should not simply be included in the v21 announcement blog post. Otherwise, looks good to me!

PR #1842, opened 2024-09-08
Update whatisvitess.md

NacerKH:
update supported versions


mattlord:
@NacerKH thank you for your interest in Vitess. However, Vitess does not support MariaDB, which is why it is not listed there. I'm going to close this for now.

PR #1843, opened 2024-09-08
Fix "apt install" command for Debian

simonhammes:
The etcd package is not available in the bookworm repositories (https://packages.debian.org/bullseye/etcd).

Related: https://github.com/vitessio/website/issues/1738

---

I only updated the docs for v20 and v21, should I also apply the changes to older versions?


mattlord (approved):
Thanks, @simonhammes ! I think that v20+ is good.

simonhammes:
@mattlord Thanks!

PR #1868, opened 2024-10-16
chore: add another video to front page

### LINKED ISSUE #1852: Show two videos on front page instead of the one we have now
It will be nice to provide more than one intro video on the front page.

thisisobate:
fixes #1852

### DIFF
--- a/layouts/partials/home/video.html
+++ b/layouts/partials/home/video.html
@@ -12,7 +12,7 @@
           class="is-video"
           src="https://www.youtube.com/embed/E6H4bgJ3Z6c?rel=0"
           allowfullscreen=""
-          style="position: relative; height: 315px; width: 560px;"
+          style="position: relative; height: 315px; width: 660px;"
           frameborder="0"
         ></iframe>
       </div>
@@ -31,7 +31,7 @@
           class="is-video"
           src="https://www.youtube.com/embed/HgSlmzC7O-E?si=E75Ot4V-pZYSN6Wy"
           allowfullscreen=""
-          style="position: relative; height: 315px; width: 560px;"
+          style="position: relative; height: 315px; width: 660px;"
           frameborder="0"
         ></iframe>
       </div>
--- a/layouts/partials/home/video.html
+++ b/layouts/partials/home/video.html
@@ -3,33 +3,24 @@
     <div class="columns is-variable is-6 is-vcentered">
       <div class="column">
         <p class="title is-size-2 is-size-3-mobile has-text-weight-light">
-          {{ T "intro_vitess" }}
+          Scaling Databases with Vitess
         </p>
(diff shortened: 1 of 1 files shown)


deepthi:
Is this done? Here's what I see on a desktop browser (Safari)
Screenshot 2024-10-17 at 11 15 04 AM
Ideally the videos would be side by side with their titles either above or below.

thisisobate:
> Is this done? Here's what I see on a desktop browser (Safari)

Yes, it's done

> Ideally the videos would be side by side with their titles either above or below

Can you explain a bit more?

nate-double-u:
> > Ideally the videos would be side by side with their titles either above or below
> 
> Can you explain a bit more?

I think something more like this:

377586231-a7204a92-5453-4d7b-8edd-666e29483b0e

frouioui (commented):
We should have the "introduction to vitess" video on the left rather than on the right. So it is the first thing the reader will see.

Nit picking, but I feel like it would look nicer if we center the videos' title in the center of their div, like:
image

Thanks for taking this on! 🚀

thisisobate:
@frouioui I already made some updates as you suggested. Feel free to take a look

layouts/partials/home/video.html
```diff
@@ -1,18 +1,27 @@
 <section class="section is-light">
   <div class="container">
-    <div class="columns is-variable is-6 is-vcentered">
-      <div class="column">
+    <div class="columns is-variable is-6 is-vcentered has-text-centered">
+      <div class="column is-6">
         <p class="title is-size-2 is-size-3-mobile has-text-weight-light">
           {{ T "intro_vitess" }}
```
frouioui, line 6: FWIW, `T "intro_vitess"` here means that we are going to rely on `en.toml` and `zh.toml` to translate either in english or chinese the title. Just a nit, but for consistency it would be nice to use the same mechanism on line 18, instead of spelling out `Scaling Databases with Vitess`, we could put it in both `en.toml` and `zh.toml`.

I am going to approve anyway as the Pull Request looks good to me, before merging I will let @deepthi 👍🏻 too as she is the author of the original issue.

frouioui (approved):

frouioui:
The Netlify preview looks really good btw! Thanks for taking this on.

nate-double-u (approved):
LGTM, thanks @thisisobate!

nate-double-u:
Oops, sorry i merged before @deepthi had a chance to 👍

Happy to help with any follow up PRs if you have comments @deepthi

layouts/partials/home/video.html
thisisobate: opened another PR for this: https://github.com/vitessio/website/pull/1877

PR #1869, opened 2024-10-16
fix loadUsers default value

rvrangel:
small change to fix the default value of `loadUsers` to `false` on the docs. this is in line [with the MySQL Shell documentation](https://dev.mysql.com/doc/mysql-shell/8.0/en/mysql-shell-utilities-load-dump.html#mysql-shell-utilities-load-dump-opt-filtering)


mattlord:
Thanks, @rvrangel ! Can you also please make the same change in the v22 docs so that it's carried forward (v22 docs will be copied to v23 when the v23 release is made and so on).

mattlord (approved):

mattlord:
Thanks again, @rvrangel ! ❤️

PR #1903, opened 2024-12-12
Operator get-stared: Fix clone and checkout commands

driv:


frouioui:
It seems like the change you have proposed does not do anything different than what we currently have. Can you explain more what issue you were facing?

I am going to close this PR for now as it's a no-op, but feel free to explain your issue.

driv:
My bad, I was following by mistake the guide for v18 and there it tries to check-out the branch befor cding.
https://vitess.io/docs/archive/18.0/get-started/operator/

It can stay closed.

PR #1906, opened 2024-12-21
Add some documentation for supported MySQL replication modes

mattlord:
Docs for https://github.com/vitessio/vitess/pull/17345

Preview: https://deploy-preview-1906--vitess.netlify.app/docs/22.0/reference/features/mysql-replication/#mysql-replication-modes

### DIFF
--- a/content/en/docs/22.0/reference/features/mysql-replication.md
+++ b/content/en/docs/22.0/reference/features/mysql-replication.md
@@ -28,9 +28,17 @@ These requirements will changed based on the durability policy.
 
 With regard to replication lag, note that this does **not** guarantee there is always at least one replica from which queries will always return up-to-date results. Semi-sync guarantees that at least one replica has the transaction in its relay log, but it has not necessarily been applied yet. The only way Vitess guarantees a fully up-to-date read is to send the request to the primary.
 
+## MySQL Replication Modes
+
+Vitess requires the use of [Row-Based Replication (RBR)](https://dev.mysql.com/doc/refman/en/replication-formats.html) with [GTIDs](https://dev.mysql.com/doc/refman/en/replication-gtids.html) enabled: [`--binlog-row-format=ROW`](https://dev.mysql.com/doc/refman/en/replication-options-binary-log.html#sysvar_binlog_format) and [`--gtid-mode=ON`](https://dev.mysql.com/doc/refman/en/replication-options-gtids.html#sysvar_gtid_mode).
+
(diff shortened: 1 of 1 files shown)


content/en/docs/22.0/reference/features/mysql-replication.md
```diff
@@ -28,9 +28,17 @@ These requirements will changed based on the durability policy.
...
 
+## MySQL Replication Modes
+
+Vitess requires the use of [Row-Based Replication (RBR)](https://dev.mysql.com/doc/refman/en/replication-formats.html) with [GTIDs](https://dev.mysql.com/doc/refman/en/replication-gtids.html) enabled: [`--binlog-row-format=ROW`](https://dev.mysql.com/doc/refman/en/replication-options-binary-log.html#sysvar_binlog_format) and [`--gtid-mode=ON`](https://dev.mysql.com/doc/refman/en/replication-options-gtids.html#sysvar_gtid_mode).
+
+Vitess also recommends [`FULL`](https://dev.mysql.com/doc/refman/en/replication-options-binary-log.html#sysvar_binlog_row_image) binary log images in order to support all manner of transformations in [VReplication workflows](../../vreplication/vreplication/), but in v17 we added *experimental* support for [`--binlog-row-image=NOBLOB`](https://dev.mysql.com/doc/refman/en/replication-options-binary-log.html#sysvar_binlog_row_image) (see [#1290](https://github.com/vitessio/vitess/pull/12905) for more details).
```
deepthi, line 35: What would it take to move NOBLOB out of "experimental"? I'm curious why this was done as experimental whereas we seem to be more confident about our ability to support PARTIAL_JSON.

deepthi (approved):

content/en/docs/22.0/reference/features/mysql-replication.md
mattlord: We expect and know that some things will fail with NOBLOB, because we lose information in the binary log events. There's no before OR after image value for the BLOB/TEXT columns when not updated in the row, which means we lose information and we know that some OnlineDDL and Materialize workflows will fail because we lack information to perform accurate transformations. We also know e.g. that changing the PK definitions on a table on the target will cause MoveTables workflows to fail if you update the PK column values in the table. That's what this is about: https://github.com/vitessio/vitess/pull/17345/files#r1892674955

With PARTIAL_JSON nothing should fail. That's because we don't lose information: there's always the BEFORE image value along with a diff in the AFTER image that can be applied to that before value in order to get the full after value.

PR #1943, opened 2025-02-20
Delete content/en/docs/21.0/reference/programs/vtbackup.md

Danmark:
Delete this duplicated file.

I'm assuming it causes a duplicated entry in the left menu and clicking either of the 'vtbackup'-links in that menu still only serves the content of the vtbackup/_index.md anyways


Danmark:
image
Looks like its working

frouioui:
This is a good point. I realized this was an issue recently, and I never got to it though. Thanks for taking a look, I'll review today 😃

Danmark:
I guess it's possible that some of the content of this file should be migrated into vtbackup/_index.md, but I'll let you be the judge of that.

frouioui:
In these `vtbackup.md` files I had done this fix: https://github.com/vitessio/website/pull/1927, which was never shown on the website as we render the `_index.md` (not the `vtbackup.md` file).

All the files in `reference/programs/` are generated using `make generated-docs` and the tool we have in `./tools/cobradocs/` - It looks at the `--help` output of every Vitess binaries and stick it in the `_index.md` files.

In order to implement https://github.com/vitessio/website/pull/1927 correctly, we would need to change the `--help` output of the `vtbackup` binary on the Vitess repository, and then re-generate the `_index.md`.

Danmark:
So, the update needs to go here https://github.com/vitessio/vitess/blob/main/go/cmd/vtbackup/cli/vtbackup.go#L141-L182

Danmark:
> So, the update needs to go here https://github.com/vitessio/vitess/blob/main/go/cmd/vtbackup/cli/vtbackup.go#L141-L182

But I guess that might ruin the formatting when viewed in terminal

frouioui (approved):
Looks good to me, thank you! Regarding https://github.com/vitessio/website/pull/1943#issuecomment-2690713763, I will see what's the best approach to fix this in our CLI.

Issue #1947, opened 2025-02-26
FAQs: Restructure FAQ Pages into Individual Pages with Short Answers


rohit-nayak-ps:
### Context

Currently, each FAQ section has a single Markdown file that contains multiple Q/As. As we add more questions, these files become larger and harder to navigate. We also have to edit existing pages each time we add or update a Q/A.

#### Per-FAQ Pages 

   - Create a separate Markdown file for each FAQ.  
   - Store the question in the page’s front matter (e.g., `question: "..."`).  
   - Put the full answer in the page’s main content.  
   - Include an optional “Short Answer” field in the front matter (e.g., `short_answer: "..."`).

#### Section Pages

   - For each FAQ section, list all questions and their Short Answers.  
   - Link each question to its dedicated page.  
   - (Optional) Provide a popup or expandable section for the full answer, but do not include it by default on the section page.

#### Hierarchy

   - Initially, keep a single-level structure similar to the current approach.  
   - If needed, we can expand to additional levels (e.g., subtopics) later.

### Benefits

- Easier navigation and maintenance: Updating or adding FAQs won’t bloat a single page.  
- Quick scanning: Users see only short answers in the section view, and can click through for full details.  
- Better organization for future expansion.

Links:
https://dariusz.wieckiewicz.org/en/adding-faq-schema-to-hugo-website/#faq-on-individual-pages

PR #1950, opened 2025-03-03
docs: Make separate md for each FAQ

beingnoble03:
- Related issue: #1947


rohit-nayak-ps:
* Let's not show each FAQ in the left bar, since that could grow really fast. So for your new link `Overview` should be the leaf. Note that we may have more sections below: for example under VReplication we may have MoveTables/Import/Reference Tables etc, each of which can have multiple FAQs. So we should only show the tree upto the parent of the FAQs. If required we can add a attribute to frontmatter for deciding this. For testing create an empty hierarchy/place holder pages. I will create some info for it once we have the formatting and current faqs in place

* On the FAQ list page (so Overview), Add an explicit link to the right of the faq title "Goto FAQ". Currently the subtitle below the title can be confused with the FAQ info:
```
What is Vitess?
Vitess is a database solution for deploying, scaling and managing large clusters of database instances.
```

* In the title for the FAQ List page, use this as the header, remove the FAQs on the next line
`FAQs: Overview` or `FAQs: VReplication/MoveTables` if there are child sections

* In the Next/Prev navigation buttons for FAQs, let's just put `Next` and `Prev` instead of the title, since those can get very long!

content/en/docs/faqv2/getting-started/overview/what-is-vitess.md
```diff
@@ -0,0 +1,12 @@
+---
+title: What is Vitess?
```
rohit-nayak-ps: Do we need both `title` and `question`? They seem to be the same value in all the faqs?

layouts/partials/docs/next-prev-navigation.html
```diff
@@ -1,22 +1,34 @@
 
     {{ if or (.PrevInSection) (.NextInSection) }}
-    <div style="padding:0;display=inline-block">
+    <div style="padding:0;">
```
beingnoble03, line 3: The CSS syntax was wrong here, we were never using `display=inline-block`

layouts/shortcodes/list-questions.html
```diff
@@ -0,0 +1,16 @@
+<div>
+  {{ range .Page.Pages }}
+    {{ $question := default .Params.title .Params.question }}
+    <h3>
+      <a href="{{ .RelPermalink }}">
+        {{ $question }}
+      </a>
+    </h3>
+    {{ $preview := .Content | plainify | htmlUnescape | safeHTML | truncate 256 }}
```
beingnoble03, line 9: It is fine to use `safeHTML` to avoid escaping of characters as these will be docs, not a user-input entity.

beingnoble03, line 9: It is fine to use `safeHTML` to avoid escaping special chars as these will be docs, not some user-generated content. More about why both `htmlUnescape` and `safeHTML` are used: [hugo docs](https://gohugo.io/functions/transform/htmlunescape/)

rohit-nayak-ps:
Currently we have both old and new versions side by side. If approved we will switch to the newer design. Hence added the `do-not-merge` label

deepthi (approved):
This is good. Let's remove the old version and merge.

beingnoble03:
Thanks @rohit-nayak-ps for pushing the change, looks all good now. ❤️

PR #1963, opened 2025-03-15
Update setup docs: Add dev.env sourcing step for all OS

mounicasruthi:
Added `source dev.env ` to the setup instructions for Ubuntu/Debian, macOS, and CentOS to ensure proper environment setup.

### DIFF
--- a/content/en/docs/contributing/build-on-macos.md
+++ b/content/en/docs/contributing/build-on-macos.md
@@ -62,7 +62,7 @@ git clone https://github.com/vitessio/vitess.git
 cd vitess
 ```
 
-Set environment variables that Vitess will require. It is recommended to put these in your `~/.bash_profile` file:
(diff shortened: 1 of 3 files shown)


content/en/docs/contributing/build-on-macos.md
```diff
@@ -69,6 +69,23 @@ Set environment variables that Vitess will require. It is recommended to put the
 export PATH=~/vitess/bin:${PATH}
 ```
 
+After setting these variables, apply the environment changes:
+
+```
+source ~/.bash_profile
```
deepthi: Given that the default shell on Mac is zsh, we should replace both references to .bash_profile with .zshrc

mounicasruthi: Done. PTAL!

GuptaManan100:
LGTM, but will leave it for deepthi to take a look and approve.

deepthi (approved):

PR #1977, opened 2025-04-09
Fix docs for `unmanaged` and `queryserver-config-max-result-size` flags

GuptaManan100:
### Description

This PR has a couple of fixes - 
1. `--unmanaged` flag: When we deprecated `--disable_active_reparents` and replaced it with `--unmanaged` flag, we made the docs changes, but didn't fix the provided example and another file. This PR fixes that.
2. `--queryserver-config-max-result-size` flag: The flag affects DMLs and the rows affected too, but the documentation was unclear regarding that.


frouioui (approved):
good catch

PR #1978, opened 2025-04-09
Delete 404ing Grafana Dashboards Page Since The Dashboard Source Does Not Exist Anymore

runewake2:
Hi, I don't see a contributing doc for this website but noticed this page was out of date and no longer functioning, apologies for any noise if you're not accepting contributions - here's a PR to delete the out-of-date doc.

This page links to the `vitess-mixin` which no longer exists: https://github.com/vitessio/vitess/tree/main/vitess-mixin. It was deleted last year as part of https://github.com/vitessio/vitess/pull/16657

A publicly available Grafana dashboard does not appear to exist in Vitess or the Website project so instead of linking to something new, this links to the deleted original dashboards instead to fix the deadlink that currently exists. I have also checked [`vitess/contrib`](https://github.com/vitessio/contrib) and do not see any dashboards there either that might be a good target for this page.

I don't know if there is additional process required here for contributing to these docs - lmk if I've missed something.

### DIFF
--- a/content/en/docs/faq/advanced-configuration/metrics/what-grafana-dashboards-are-available.md
+++ b/content/en/docs/faq/advanced-configuration/metrics/what-grafana-dashboards-are-available.md
@@ -5,4 +5,4 @@ notoc: true
 weight: 1
 ---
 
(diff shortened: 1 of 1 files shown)


frouioui (commented):
I think we should probably edit the page, instead of removing it, so we can link to the original PR, explaining it got removed in https://github.com/vitessio/vitess/pull/16657. So that even if we still don't have the source code, people can access the information one way.

runewake2:
Thanks! Let me add the doc back with the links. That seems like a useful way to preserve this information.

frouioui (approved):

frouioui:
Sorry for the delay! Lgtm!

PR #1998, opened 2025-07-25
Update are-foreign-keys-supported-in-vitess.md

nithinsag:
FIx typo


harshit-gangal:
We do not take in typo fix PRs unless there is some concrete changes that have typo changes along with it.

PR #2001, opened 2025-08-25
Add docs for new `transaction_timeout` system variable

mhamza15:
Add docs for new `transaction_timeout` system variable introduced in https://github.com/vitessio/vitess/pull/18560.


harshit-gangal:
We should add a section for transaction as well in `reference/compatibility/mysql-compatibility`

maxenglander (approved):

PR #2009, opened 2025-10-12
Fix `MirrorTraffic` command output

mhamza15:
Small change to fix the output for `MirrorTraffic` in the `MoveTables` example docs.


mattlord (approved):
Thanks, @mhamza15 ! ❤️

PR #2025, opened 2025-11-14
add docs for --init-tablet-type-lookup + remove PRIMARY from --init-tablet-type

sbaker617:
requested for this [PR](https://github.com/vitessio/vitess/pull/18777):

adds docs for new flag: `--init-tablet-type-lookup`

remove `PRIMARY` as valid type for `--init-tablet-type`


mattlord (approved):
❤️

PR #2028, opened 2025-11-26
Add docs for TransactionChunkSize

twthorn:
Add docs for https://github.com/vitessio/vitess/pull/18849

### DIFF
--- a/content/en/docs/24.0/reference/vreplication/vstream.md
+++ b/content/en/docs/24.0/reference/vreplication/vstream.md
@@ -154,6 +154,13 @@ response and stop sending any further events.
 
 This replaces the `in_order` hint (e.g. `"in_order:REPLICA,PRIMARY"`) previously used to specify tablet type order [during source tablet selection](../tablet_selection/).
 
+##### TransactionChunkSize
+
+**Type** int64 (bytes)\
+**Default** 134217728 (128MB)
+
+Transaction chunk threshold in bytes. When a transaction exceeds this size, VTGate will acquire a lock to ensure contiguous, non-interleaved delivery (BEGIN...ROW...COMMIT sent sequentially without mixing events from other shards). Events are still chunked to prevent OOM. Transactions smaller than this threshold are sent without locking for better parallelism.
+
 ### RPC Response
 
 The [`VStream` gRPC](https://pkg.go.dev/vitess.io/vitess/go/vt/vtgate/vtgateconn#VTGateConn.VStream) returns
--- a/content/en/docs/24.0/reference/vreplication/vstream.md
+++ b/content/en/docs/24.0/reference/vreplication/vstream.md
@@ -168,6 +168,10 @@ transaction in memory before sending to the client. This prevents out-of-memory 
 Transactions smaller than this threshold are sent without locking for better parallelism across shards. The default
 128MB works for most use cases, but you may want to adjust it based on your expected transaction sizes and available memory.
 
+{{< warning >}}
+This flag will be ignored if the `MinimizeSkew` flag is also specified because the combination of these two features can lead to deadlocks.
+{{< /warning >}}
+
 ### RPC Response
 
 The [`VStream` gRPC](https://pkg.go.dev/vitess.io/vitess/go/vt/vtgate/vtgateconn#VTGateConn.VStream) returns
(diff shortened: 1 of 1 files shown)


content/en/docs/24.0/reference/vreplication/vstream.md
```diff
@@ -154,6 +154,13 @@ response and stop sending any further events.
 
 This replaces the `in_order` hint (e.g. `"in_order:REPLICA,PRIMARY"`) previously used to specify tablet type order [during source tablet selection](../tablet_selection/).
 
+##### TransactionChunkSize
+
+**Type** int64 (bytes)\
+**Default** 134217728 (128MB)
+
+Transaction chunk threshold in bytes. When a transaction exceeds this size, VTGate will acquire a lock to ensure contiguous, non-interleaved delivery (BEGIN...ROW...COMMIT sent sequentially without mixing events from other shards). Events are still chunked to prevent OOM. Transactions smaller than this threshold are sent without locking for better parallelism.
```
mattlord:
```suggestion
Transaction chunking threshold in bytes. When a transaction exceeds this size, VTGate will acquire a lock to ensure contiguous, non-interleaved delivery:  BEGIN...ROW...COMMIT sent sequentially without mixing events from other shards. Events within the transaction are sent in chunks of this size in order to prevent having to process the entire transaction in memory at once, resulting in excessive memory usage and risking an OOM event. Transactions smaller than this threshold are processed as a single batch, entirely in memory, and sent as a single message (and without locking for better parallelism).
```

content/en/docs/24.0/reference/vreplication/vstream.md
```diff
@@ -154,6 +154,13 @@ response and stop sending any further events.
 
 This replaces the `in_order` hint (e.g. `"in_order:REPLICA,PRIMARY"`) previously used to specify tablet type order [during source tablet selection](../tablet_selection/).
 
+##### TransactionChunkSize
+
+**Type** int64 (bytes)\
+**Default** 134217728 (128MB)
+
+Transaction chunking threshold in bytes. When a transaction exceeds this size, VTGate will acquire a lock to ensure contiguous, non-interleaved delivery:  BEGIN...ROW...COMMIT sent sequentially without mixing events from other shards. Events within the transaction are sent in chunks of this size in order to prevent having to process the entire transaction in memory at once, resulting in excessive memory usage and risking an OOM event. Transactions smaller than this threshold are processed as a single batch, entirely in memory, and sent as a single message (and without locking for better parallelism).
```
mattlord: FWIW, here was promptless's suggestion
```
##### TransactionChunkSize

**Type** int64 (bytes)\
**Default** 134217728 (128MB)

When a transaction exceeds this size threshold, VTGate acquires a lock to ensure contiguous, non-interleaved delivery of the transaction's events. All events for the transaction (BEGIN...ROW...COMMIT) are sent sequentially without mixing in events from other shards.

By setting this threshold, VTGate can stream large transaction events in chunks rather than buffering the entire transaction in memory before sending to the client. This prevents out-of-memory (OOM) errors while maintaining transaction integrity and ordering guarantees.

Transactions smaller than this threshold are sent without locking for better parallelism across shards. The default 128MB works for most use cases, but you may want to adjust it based on your expected transaction sizes and available memory.
```

mattlord: I think I like that better 🙂

twthorn: Thanks, updated!

mattlord (approved):

content/en/docs/24.0/reference/vreplication/vstream.md
```diff
@@ -154,6 +154,20 @@ response and stop sending any further events.
...
+
+If the size threshold is <= 0, transaction chunking is disabled. Clients must explicitly set this to opt-in to transaction chunking.
+
+If enabled, when a transaction exceeds this size threshold, VTGate acquires a lock to ensure contiguous, non-interleaved delivery
+of the transaction's events. All events for the transaction (BEGIN...ROW...COMMIT) are sent sequentially without mixing in events from other shards.
+By setting this threshold, VTGate can stream large transaction events in chunks rather than buffering the entire
+transaction in memory before sending to the client. This prevents out-of-memory (OOM) errors while maintaining transaction integrity and ordering guarantees.
+Transactions smaller than this threshold are sent without locking for better parallelism across shards. The default
+128MB works for most use cases, but you may want to adjust it based on your expected transaction sizes and available memory.
+
```
mattlord, line 170:
```suggestion

{{< warning >}}
This flag will be ignored if the `MinimizeSkew` flag is also specified because the combination of these two features can lead to deadlocks.
{{< /warning >}}

```

twthorn:
+1 thanks for adding minimize skew caveat

PR #2055, opened 2026-01-13
Xtrabackup configuration instructions

itmisx:
XtraBackup doesn't support hostPath when not a minikube cluster.Because shared storage is required, hostpath cannot share storage when deployed across multiple nodes.

### DIFF
--- a/content/en/docs/23.0/get-started/operator.md
+++ b/content/en/docs/23.0/get-started/operator.md
@@ -69,7 +69,7 @@ In this directory, you will see a group of yaml files. The first digit of each f
 kubectl apply -f 101_initial_cluster.yaml
 ```
 
-{{<info>}} If it's not a minikube cluster, and the backup engine is xtrabackup, please configure the locations to file, gcs, s3, ceph, etc., or customize pvc.{{<info>}}
+{{<info>}} If it's not a minikube cluster, and the backup engine is xtrabackup, please configure the locations to file, gcs, s3, ceph, etc., or customize pvc. {{</info>}}
 
 for example
 ```yaml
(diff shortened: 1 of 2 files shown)


mattlord:
Thanks, @itmisx ! 

This is the preview: https://deploy-preview-2055--vitess.netlify.app/docs/23.0/get-started/operator/#bring-up-an-initial-cluster

The formatting is a bit off with the brackets. It probably makes sense to put the new section inside of an info block:
```
{{< info >}}
...
{{< /info >}}
```

Maybe that's what you meant to do?

Whenever we're happy with the content, would you mind copying it to the 24.0 page as well? Then the changes will be carried forward as the 24.0 docs will be copied to the 25.0 docs when the 24.0 RC is released and `main` becomes 25.0.

Thanks again! ❤️

content/en/docs/23.0/get-started/operator.md
```diff
@@ -68,7 +68,8 @@ In this directory, you will see a group of yaml files. The first digit of each f
 ```bash
 kubectl apply -f 101_initial_cluster.yaml
 ```
-{{}}If it's not a minikube cluster, and the backup engine is xtrabackup, please configure the locations to file, gcs, s3, ceph, etc., or customize pvc.{{}}
+
+{{<info>}} If it's not a minikube cluster, and the backup engine is xtrabackup, please configure the locations to file, gcs, s3, ceph, etc., or customize pvc.{{<info>}}
```
mattlord: I think that we should end the info block (`{{</ info >}}`) after the example (line 83) as well since all of this content is related.

itmisx: I've made the changes, sorry, I'm not very familiar with rendering syntax.

mattlord (approved):
Thanks, @itmisx !

PR #2081, opened 2026-03-05
fix: update teardown script reference from 401 to 501 for v22+

khkim6040:
## What's this?

The teardown script was renamed from `401_teardown.sh` to `501_teardown.sh` in vitessio/vitess#16893 when backup/restore scripts (40x) were added. This was included starting from release-22.0, but the website docs still reference the old name.

## Changes

- Updated `401_teardown.sh` → `501_teardown.sh` in v22.0, v23.0, v24.0 docs (en/zh)
- Updated contributing guides (build-on-centos, build-on-macos, build-on-ubuntu)
- Left v21.0 and older docs unchanged since they correctly reference `401_teardown.sh`

15 files changed, 21 substitutions total.

### DIFF
--- a/content/en/docs/22.0/get-started/local-mac.md
+++ b/content/en/docs/22.0/get-started/local-mac.md
@@ -296,10 +296,10 @@ The schema has been simplified to include only those fields that are significant
 
 You can now proceed with [MoveTables](../../user-guides/migration/move-tables).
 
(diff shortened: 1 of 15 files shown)


mattlord (approved):

mattlord:
Thanks, @khkim6040 ! ❤️

PR #2101, opened 2026-04-24
Add vstream max age docs and some missing flags

twthorn:
We recently added a new flag so add docs for that
https://github.com/vitessio/vitess/pull/19800

Also add docs for some other new-ish flags that missed being documented,.


mattlord (approved):
Thanks, @twthorn ! ❤️

PR #2106, opened 2026-04-30
Adjust archived docs structure

mhamza15:
When we archive older versions, we move their docs from `/docs/<version>` to `/docs/archive/<version>`. This can break links and hurt SEO. This makes a few changes:

- Moves all archived docs from `/docs/archive/<version>` back to `/docs/<version>`
- Set up a redirect from `/docs/archive/<version>` to `/docs/<version>` to maintain existing links
- Added an `archived` boolean attribute, true for all archived versions
- Archived versions are filtered from the sidebar, and have a callout banner linking to the current version
- Updates `current` to v24 and `next` to v25, which was missed as part of the recent v24 release

image


mattlord:
This was addressed in https://github.com/vitessio/website/pull/2107

PR #2125, opened 2026-05-24
Document `--force-cut-over-after` immediate behavior for sub-1ms values

shlomi-noach:
Clarify in `v23`, `v24`, and `v25` (current) docs that any positive duration of `1ms` or less forces the cut-over starting with the very first attempt, and that the value must be strictly greater than zero.

This behavior was introduced in https://github.com/vitessio/vitess/pull/18423 and we did not document it at the time.


mattlord (approved):
❤️

PR #2137, opened 2026-06-05
docs: rename programs references to cli reference

voidgremlin19:
## Summary

- Rename references from `reference/programs` to `reference/cli`
- Update documentation links to match the new CLI reference structure
- Improve consistency across documentation

## Testing

- Verified updated documentation links
- Reviewed affected documentation files


mattlord:
Why? This is not something we wanted to do and it will break a lot of existing links unless we also add more redirects/aliases. But cli isn't correct either, for all of the various things we have in that section. I'm going to close this for now, if I'm missing something here please let me know.

PR #2149, opened 2026-07-12
docs: fix link to Keyspace ID in vindexes documentation

obsfx:
Currently the `Keyspace ID` link we have in the `VIndexes` document is leading to a 404 page

image

This PR is just about to fix it by removing the unnecessary `.md` suffix at the end of the link


mattlord (approved):
Thanks, @obsfx ! We should also change this in the 23.0 and 25.0 docs. The 25.0 docs will become the 26.0 docs and so on so at a minimum we should do it there as well.

Thanks!

obsfx:
Hi @mattlord thanks for pointing out that versions as well. I also updated the links for the `v23.0` and `v25.0`🙌

mattlord (approved):

